PR #11, opened 2018-06-08
Fix extra = sign on type declaration and indentation on case statement

rogeriochaves:
Hello,

I'm using this lib to write an elm code generator, and I found a few bugs as you can see on the gif below:

new route generator

The changes fixes those bugs


stil4m:
Thanks @rogeriochaves. 👍 

Will release this with a few other bug fixes soon.

PR #13, opened 2018-06-09
Fix qualified expressions

rogeriochaves:
It was missing a dot, qualified expressions were outputting like this:

```elm
Htmlmap
```

now it outputs correctly:

```elm
Html.map
```


stil4m:
Thanks :)

PR #15, opened 2018-06-10
Fix writing subfunctions type signatures

rogeriochaves:
The writing was outputting `a -> b -> c` instead of `(a -> b) -> c`


stil4m:
Nice. Thanks!

PR #18, opened 2018-07-03
Allow empty record expressions

zwilias:
Record expressions like `{}` are valid Elm, but trip up the parser.

(I'm lazy so it's based on #17 but feel free to rebase!)


stil4m:
Thanks for the great work Ilias 👍.

PR #19, opened 2018-07-08
Encode `StringPattern` using `"string"`

zwilias:
As this is what the decoder expects, too.


stil4m:
@zwilias great! Thanks :)

stil4m:
Just released `4.0.4` with this fix.

PR #33, opened 2019-10-08
Huge case statements

zwilias:
Attempting to parse very (VERY) large union types and functions using them currently results in the stack being blown.

We're encountering this in our codebase in the wild, so this isn't a purely theoretical thing.

This program prints out a module that should crash `elm-syntax`/`elm-analyse` as it currently exists: 

```elm
module Main exposing (main)

import Browser
import Html exposing (Html, pre, text)


ctors =
    List.range 1 2000
        |> List.map (\x -> "Ctor" ++ String.fromInt x)


decl =
    """
type MyType
  = """
        ++ String.join "\n  | " ctors


use =
    """
myFun : MyType -> String
myFun foo =
  case foo of
""" ++ String.join "\n" (List.map (\x -> "    " ++ x ++ " -> \"" ++ x ++ "\"") ctors)


main =
    pre []
        [ text ("module Main exposing (..)\n\n" ++ decl ++ "\n\n" ++ use) ]
```

I didn't really see a better way to handle the case-statements than having it use `loop` as well, so I added a `Combine.loop` helper for that. No pressure to actually use this suggestion, though - I'm sure alternatives are possible! The simplification of `valueConstructors` using `sepBy1` looks nice, though. Hope I didn't miss anything there!

### DIFF
--- a/demo/elm.json
+++ b/demo/elm.json
@@ -7,7 +7,6 @@
     "elm-version": "0.19.0",
     "dependencies": {
         "direct": {
-            "Microsoft/elm-json-tree-view": "2.0.0",
             "elm/browser": "1.0.0",
             "elm/core": "1.0.0",
             "elm/html": "1.0.0",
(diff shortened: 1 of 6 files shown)


stil4m:
@zwilias Nice! Great work on this :).

Issue #41, opened 2020-04-24
Operator precedence is not taken into account when parsing expressions


jhrcek:
It seems to me, that operator precedence is not take into account when parsing expressions.

For example the following two expressions
```elm
1 ^ 2 * 3 + 4
1 + 2 * 3 ^ 4
```

both produce the same shape of the expression tree, the first one being obviously incorrect

```
^
│
├─ 1
│
└─ *
   │
   ├─ 2
   │
   └─ +
      │
      ├─ 3
      │
      └─ 4

+
│
├─ 1
│
└─ *
   │
   ├─ 2
   │
   └─ ^
      │
      ├─ 3
      │
      └─ 4
```

I created SSCCE which demonstrated this issue: https://github.com/jhrcek/elm-syntax-sscce

Wider context: this library is used by [elm-review](https://github.com/jfmengels/elm-review), which -among other things - allows writing custom linting rules for elm source code. Unfortunately since the expressions are not parsed correctly by this library, this leads to all sorts of false positives as in the following example

![Screenshot from 2020-04-24 09-34-06](https://user-images.githubusercontent.com/2716069/80186747-cb7f4080-860e-11ea-984c-3d0fd7cb66c7.png)

Because elm-syntax parses this as `(f <| x) || y` it gives this false positive warning, which shouldn't be there, because `||` has higher precedence than `<|`, so the correct parsing should look like `f <| (x || y)`.

MartinSStewart:
@stil4m I think this issue can be closed now.

stil4m:
@jhrcek Thank you for your report. That was clearly a bug.
The fix @MartinSStewart made a functional fix for this which is already released in 7.1.2.

I've just prepared a pull request (https://github.com/stil4m/elm-syntax/pull/66) that will ship with 8.0.0, which will ensure that most programs have proper parsing of operators (and will nog require parsing of dependencies).

jhrcek:
Thank you both @MartinSStewart and @stil4m , I appreciate your work :heart:

PR #52, opened 2020-06-14
Fix operator precedence

MartinSStewart:
This should fix https://github.com/stil4m/elm-syntax/issues/41

As far as I can tell, two problems occurred. One was a bug in elm-syntax and one was a mistake in the bug report. 

The bug in elm-syntax (which this PR fixes) is that operator precedence is handled in reverse order. So in `1 * 2 + 3`, elm-syntax would incorrectly parse it as `1 * (2 + 3)`. It should now correctly parse it as `(1 * 2) + 3`

The mistake with the bug report was that, unless the operator definitions are also parsed, elm-syntax treats all operators with the same precedence. This is why, in the bug report SSCCE, elm-syntax appeared to have no handling for operator precedence and would just nest operators left to right.

@stil4m Assuming I've gotten all this right, can this be made as a patch release? (as opposed to getting bundled with breaking changes and waiting until a major release)

### DIFF
--- a/src/Elm/Processing.elm
+++ b/src/Elm/Processing.elm
@@ -228,12 +228,12 @@ findNextSplit dict exps =
 lowestPrecedence : List ( String, Infix ) -> Dict String Infix
 lowestPrecedence input =
     let
-        maxi =
+        mini =
             input
                 |> List.map (Tuple.second >> .precedence >> Node.value)
                 |> List.minimum
     in
-    maxi
+    mini
         |> Maybe.map (\m -> List.filter (Tuple.second >> .precedence >> Node.value >> (==) m) input)
         |> Maybe.withDefault []
         |> Dict.fromList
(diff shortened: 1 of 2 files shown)


src/Elm/Processing.elm
```diff
@@ -225,13 +225,13 @@ findNextSplit dict exps =
...
+lowestPrecedence : List ( String, Infix ) -> Dict String Infix
+lowestPrecedence input =
     let
         maxi =
             input
                 |> List.map (Tuple.second >> .precedence >> Node.value)
-                |> maximum
+                |> List.minimum
     in
     maxi
```
stil4m, line 236: `maxi` seems to be an incorrect variable name in this context. Could you rename `maxi` to `minimumValue`? Or maybe better, just eliminate the `let-in` in the definition of `lowestPrecedence` (`maxi` is only used once as far as I can see).

MartinSStewart:
This pull request was duplicated and the duplicate was merged instead. This can also be closed.

Issue #43, opened 2020-05-07
Use nonempty list to make impossible syntax impossible


MartinSStewart:
There are a few places in elm-syntax where  using `List.Nonempty` instead of `List` is used when could prevent needing to deal with impossible cases.

The ones I've been able to find are 

* `Elm.Syntax.Expression.RecordUpdateExpression` - Having no record setters would correspond to `{ a | }` which is invalid syntax.
* `Elm.Syntax.Expression.Application` - Not sure how to give an example of this one. If the list is empty then there's just nothing there right?
* `Elm.Syntax.ModuleName.ModuleName` - Here I can understand why it's like this. `Elm.Syntax.Expression.FunctionOrValue` can have an empty ModuleName because functions can be local or exposed. However, in import statements, having a ModuleName that's just an empty list would look like this `import ` which is invalid syntax. One solution is to have to two versions of ModuleName, one is a list, one is a nonempty list. Another solution is for FunctionOrValue to treat the last element in the nonempty list as the function/value name.

MartinSStewart:
I realized that using List.Nonempty isn't a good idea since it would prevent pattern matching on the result. `(a, List a)` is probably the way to go.

jhrcek:
You could also consider using https://package.elm-lang.org/packages/turboMaCk/non-empty-list-alias/latest/ which uses `(a, List a)` to represent nonempty list and just adds functions for working with that type.

jfmengels:
@MartinSStewart this can be closed now, right?

MartinSStewart:
I only made a pull request for this bullet point `Elm.Syntax.Expression.Application`. The rest hasn't been addressed.

jfmengels:
Ah yes, my bad :+1:

jfmengels:
For `FunctionOrValue`, the module name could also be `Maybe ModuleName`, which i think would remove the need for a separate potentially empty module name.

MartinSStewart:
I spoke with @stil4m on Slack about what type signature to use for nonempty lists. We settled on `Application (Node Expression) (List (Node Expression))` rather than use `Application (Node Expression, List (Node Expression))` or `List.Nonempty`.

The reasoning for this is that it reduces how much syntax is needed when doing pattern matching. This comes at the expense of making it a little harder to use https://package.elm-lang.org/packages/turboMaCk/non-empty-list-alias/latest/ but we decided this tradeoff is worthwhile as the user will likely be doing a lot of pattern matching and deconstructing expressions.

Issue #49, opened 2020-06-12
Simplify TypeAnnotation 


jfmengels:
I think the [`TypeAnnotation`](https://package.elm-lang.org/packages/stil4m/elm-syntax/latest/Elm-Syntax-TypeAnnotation#TypeAnnotation) type could be simplified a bit.

[`elm/project-metadata-utils`](https://package.elm-lang.org/packages/elm/project-metadata-utils/1.0.1/Elm-Type#Type) provides a different representation for the same data (without the `Range`s).

In the following proposals, I will have a table of equivalences, where `<thing>` means that something is wrapped in a `Node`.

If you prefer discussing each one in different issues, let me know and I'll be happy to create them.

### Removing `Unit` in favor of `Tupled []`

| Elm code |elm-syntax|elm/project-metadata-utils|
| ------ | ------------- | ------------- |
|()| Unit | Tuple [] |
|( a, b )| Tupled [ <GenericType "a">, <GenericType "b">  ] | Tuple [ Var "a", Var "b" ] |

I think that the `Unit` serves little purpose, and could be simplified to `Tupled []`, like what `project-metadata-utils` does.

Changing this doesn't make pattern matching harder, as you can still pattern match for `Tupled []`.

### Fusing GenericRecord and Record

| Elm code |elm-syntax|elm/project-metadata-utils|
| ------ | ------------- | ------------- |
| { a : () } | Record [ <( <"a">, \<Unit> )> ] | Record [ ( "a", Tuple [] ) ] Nothing |
| { b \| a : () } | GenericRecord <"b"> [ <( <"a">, \<Unit> )> ] | Record [ ( "a", Tuple [] ) ] (Just "b") |

`elm/project-metadata-utils` employs the same type for records and generic records, and uses a `Maybe` for the optional generic part.
Both Record and GenericRecord serve similar purposes, and in practice I think that we will analyse them in similar ways. Having a single type can help avoid duplication of code.

Again, this does not prevent pattern matching directly on generic and non-generic record, as you can still pattern match for `Record _ Nothing` and `Record _ (Just _)`

Changing this doesn't make pattern matching harder, as you can still pattern match for `Tupled []`.

### Renaming `Tupled` to `Tuple` and `Typed` to `Type`

| Elm code |elm-syntax|elm/project-metadata-utils|
| ------ | ------------- | ------------- |
| a | GenericType "a"|Var "a"|
[...]

PR #55, opened 2020-06-15
Make sure RecordUpdateExpression has at least one field 

MartinSStewart:
I've changed `RecordUpdateExpression (Node String) (List (Node RecordSetter))` to `RecordUpdateExpression (Node String) (Node RecordSetter) (List (Node RecordSetter))` to make sure it's not possible to have a record update with no fields.

This fixes the first bullet point in this issue https://github.com/stil4m/elm-syntax/issues/43

### DIFF
--- a/src/Elm/Inspector.elm
+++ b/src/Elm/Inspector.elm
@@ -371,10 +371,10 @@ inspectInnerExpression config expression context =
         RecordExpr expressionStringList ->
             List.foldl (\a b -> inspectExpression config (Tuple.second <| Node.value a) b) context expressionStringList
 
-        RecordUpdateExpression name updates ->
+        RecordUpdateExpression name firstUpdate updates ->
             actionLambda config.onRecordUpdate
-                (\c -> List.foldl (\a b -> inspectExpression config (Tuple.second <| Node.value a) b) c updates)
-                ( name, updates )
+                (\c -> List.foldl (\a b -> inspectExpression config (Tuple.second <| Node.value a) b) c (firstUpdate :: updates))
+                ( name, firstUpdate :: updates )
                 context
 
 
--- a/src/Elm/Inspector.elm
+++ b/src/Elm/Inspector.elm
@@ -46,7 +46,7 @@ type alias Config context =
     , onCase : Order context Case
     , onFunctionOrValue : Order context ( ModuleName, String )
     , onRecordAccess : Order context ( Node Expression, Node String )
-    , onRecordUpdate : Order context ( Node String, List (Node RecordSetter) )
+    , onRecordUpdate : Order context ( Node String, Node RecordSetter, List (Node RecordSetter) )
     }
 
 
@@ -374,7 +374,7 @@ inspectInnerExpression config expression context =
         RecordUpdateExpression name firstUpdate updates ->
             actionLambda config.onRecordUpdate
                 (\c -> List.foldl (\a b -> inspectExpression config (Tuple.second <| Node.value a) b) c (firstUpdate :: updates))
-                ( name, firstUpdate :: updates )
+                ( name, firstUpdate, updates )
                 context
 
 
--- a/src/Elm/Syntax/Expression.elm
+++ b/src/Elm/Syntax/Expression.elm
@@ -106,7 +106,7 @@ type Expression
     | ListExpr (List (Node Expression))
     | RecordAccess (Node Expression) (Node String)
     | RecordAccessFunction String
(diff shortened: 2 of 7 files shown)


src/Elm/Inspector.elm
```diff
@@ -371,10 +371,10 @@ inspectInnerExpression config expression context =
         RecordExpr expressionStringList ->
             List.foldl (\a b -> inspectExpression config (Tuple.second <| Node.value a) b) context expressionStringList
 
-        RecordUpdateExpression name updates ->
+        RecordUpdateExpression name firstUpdate updates ->
             actionLambda config.onRecordUpdate
```
stil4m, line 375: I think the signature of `onRecordUpdate` should be expanded. We have the guarantee that there is a non-empty list, let's use it there as well (`( Node String, List (Node RecordSetter) )` -> `( Node String, Node RecordSetter, List (Node RecordSetter) )`. In that case `( name, firstUpdate :: updates )` would become `( name, firstUpdate,  updates )`

src/Elm/Syntax/Expression.elm
```diff
@@ -333,11 +333,11 @@ encodeLetBlock { declarations, expression } =
...
 
 
-encodeRecordUpdate : Node String -> List (Node RecordSetter) -> Value
-encodeRecordUpdate name updates =
+encodeRecordUpdate : Node String -> Node RecordSetter -> List (Node RecordSetter) -> Value
+encodeRecordUpdate name firstUpdate updates =
     JE.object
         [ ( "name", Node.encode JE.string name )
-        , ( "updates", JE.list (Node.encode encodeRecordSetter) updates )
+        , ( "updates", JE.list (Node.encode encodeRecordSetter) (firstUpdate :: updates) )
```
stil4m, line 340: I think it is better to write to a different JSON structure as well. Maybe add a JSON field `firstUpdate`? This makes the decoder a bit more simpler as well.

MartinSStewart: I did it this way because it doesn't break existing JSON encoded data.

src/Elm/Inspector.elm
MartinSStewart: Okay, I think I've fixed it. Let me know if I've missed something.

src/Elm/Syntax/Expression.elm
jfmengels: Maybe this can be simplified for the next major version?

stil4m: @MartinSStewart very good that you thought about the compatibility of the JSON structure. Some may have overseen this.

Like @jfmengels notes, the next version will be a breaking change. There is already stuff on master that indicates this, and I'm working on #49 that will also break the JSON.

I think that the JSON interface compatibility should comply with the SemVer rules of the package itself. In that case I find it justified to break compatibility. What do you think?

jfmengels: I think it makes sense, but it's best to warn about that in the documentation.

At the moment in `elm-review`, I cache every AST on the file-system (because parsing is very expensive), and I load them.
The cache is agnostic of the version of `elm-syntax`, so this means that potentially, the AST will become invalid if the version for `elm-syntax` gets bumped.

That said, if the decoding of the AST fails, then we fallback to simply parsing the file again. We don't lose much except for time on the initial run after `elm-syntax` has been bumped. So either I accept this trade-off, or I store the ASTs under a folder for the "current" version of `elm-syntax` (which I need to figure out, but in my case that is easy).

I mainly mention this for context, since `elm-review` will not crash and it's in practice not a big problem. But for other applications this might be a problem, as you need to account for the AST being potentially incorrect.

Let's discuss this in a different issue if you wish to, in order not to derail this discussion even more.

stil4m: I think what @jfmengels does in `elm-review` should be how everybody should use the JSON format. It is more for caching and such than a 'communication' interface. I'll create a story to make this clear in the README/docs. For now @MartinSStewart it may be wise to just create a breaking change. It keeps the code more clean.

MartinSStewart: @stil4m I wrote a response in this issue if you have a moment https://github.com/stil4m/elm-syntax/issues/56

stil4m: Lets merge this. Then I can resolve potential conflicts with #59. #56 will be an improvement for a later stage (outside of this PR).

PR #64, opened 2020-06-17
Compare Ranges

sparksp:
This adds tests for `Range.combine` and adds new comparison functions:

* `Range.compare : Range -> Range -> Order`
* `Range.compareLocations : Range.Location -> Range.Location -> Order`

~~After a discussion with @stil4m we settled on `Range.comparePositions` rather than simply `Range.compare` so that there's no confusion about what dimension is being compared (i.e., position or size). There's no easy way to get the size of a range so that comparison is not provided.~~

### DIFF
--- a/src/Elm/Syntax/Range.elm
+++ b/src/Elm/Syntax/Range.elm
@@ -1,7 +1,7 @@
 module Elm.Syntax.Range exposing
     ( Range, Location
     , emptyRange, combine
-    , compareLocations, comparePositions
+    , compare, compareLocations
     , encode, decoder
     )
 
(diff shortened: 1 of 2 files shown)


jfmengels:
For me, a "location" is the same thing as a "position". I would argue that `compareRanges` would be less confusing, even if feels redundant.

Also, I am not sure that anybody would be interested in comparing range sizes, so I don't know how confusing this will be in practice, so I am not sure the `Ranges` is even necessary.

I would argue that this might be nicer:
```elm
Range.compare : Range -> Range -> Order
Range.compareLocations : Range.Location -> Range.Location -> Order
Range.compareRangeSizes : Range -> Range -> Order -- if that ever gets added
```

Anyway, `Position` is very confusing to me, so `compare` or `compareRanges` would be an improvement to the naming in my opinion.

sparksp:
AppVeyor is failing because of an upstream change...

Messages:
- src\Elm\Parser\Declarations.elm
  > Unused imported variable `Location` at ((21,44),(21,52))

stil4m:
Just fixed the CI. Could you rebase?

sparksp:
Rebased and all green now ✅

PR #71, opened 2020-06-19
Require at least one case statement to appear in case block

MartinSStewart:
This prevents a case block from having 0 cases which should be impossible.


src/Elm/Syntax/Expression.elm
```diff
@@ -143,7 +143,8 @@ type alias Lambda =
 -}
 type alias CaseBlock =
     { expression : Node Expression
-    , cases : Cases
+    , firstCase : Case
+    , restOfCases : List Case
```
MartinSStewart, line 147: Initially I had `cases : (Case, List Case)` but without requiring https://package.elm-lang.org/packages/turboMaCk/non-empty-list-alias/latest/ it felt awkward to use so I split it into two record fields instead.

src/Elm/Syntax/Expression.elm
```diff
@@ -153,12 +154,6 @@ type alias Case =
     ( Node Pattern, Node Expression )
 
 
-{-| Type alias for a list of cases
--}
-type alias Cases =
-    List Case
```
MartinSStewart, line 159: I've removed `type alias Cases` as it felt like necessary indirection.

PR #72, opened 2020-06-20
Require at least one exposed item in an explicit exposing list

MartinSStewart:
This prevents another impossible state

### DIFF
--- a/src/Combine.elm
+++ b/src/Combine.elm
@@ -231,12 +231,12 @@ many1 p =
 
 sepBy : Parser s x -> Parser s a -> Parser s (List a)
 sepBy sep p =
-    or (sepBy1 sep p) (succeed [])
+    or (sepBy1 sep p |> map (\( head, rest ) -> head :: rest)) (succeed [])
 
 
-sepBy1 : Parser s x -> Parser s a -> Parser s (List a)
+sepBy1 : Parser s x -> Parser s a -> Parser s ( a, List a )
 sepBy1 sep p =
-    succeed (::)
+    succeed Tuple.pair
         |> andMap p
         |> andMap (many (sep |> continueWith p))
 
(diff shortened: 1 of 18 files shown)


src/Combine.elm
```diff
@@ -231,12 +231,12 @@ many1 p =
 
 sepBy : Parser s x -> Parser s a -> Parser s (List a)
 sepBy sep p =
-    or (sepBy1 sep p) (succeed [])
+    or (sepBy1 sep p |> map (\( head, rest ) -> head :: rest)) (succeed [])
 
 
-sepBy1 : Parser s x -> Parser s a -> Parser s (List a)
+sepBy1 : Parser s x -> Parser s a -> Parser s ( a, List a )
```
MartinSStewart, line 237: I've modified sepBy1 to return a nonempty list rather than lose this information. This has helped make it clear which other datastructures could also use nonempty lists.

src/Elm/Syntax/Exposing.elm
```diff
@@ -134,10 +134,24 @@ decoder : Decoder Exposing
 decoder =
     decodeTyped
         [ ( "all", Range.decoder |> JD.map All )
-        , ( "explicit", JD.list topLevelExposeDecoder |> JD.map Explicit )
+        , ( "explicit", decodeNonemptyList topLevelExposeDecoder |> JD.map (\( head, rest ) -> Explicit head rest) )
         ]
 
 
+decodeNonemptyList : Decoder a -> Decoder ( a, List a )
```
MartinSStewart, line 141: This function is duplicated from Expression.elm since I didn't want to expose that function to users. I'm guessing this all will get removed anyway once we start using elm-codec.

tests/tests/Elm/Parser/Samples.elm
```diff
@@ -422,7 +422,7 @@ foo = 1
 
 sample4 : String
 sample4 =
-    """module Operators exposing ()
+    """module Operators exposing (..)
```
MartinSStewart, line 425: I believe the invalid exposing here was a mistake so I've fixed it.

Issue #76, opened 2020-06-24
Use a variation of Pattern for function arguments


MartinSStewart:
Currently, the arguments used in a function declaration are `List Pattern`. The problem is that there are many variants of Pattern that aren't valid when deconstructing an argument.

For example:
```elm
-- Here we have the IntPattern variant. This isn't valid syntax.
myFunction 2 = 
    ...
```

As far as I can tell, only these patterns variants are valid
```
AllPattern
UnitPattern
TuplePattern
RecordPattern
VarPattern
NamedPattern
AsPattern
ParenthesizedPattern
``` 
I think it makes sense then to have a new custom type (`ArgumentPattern` perhaps) that contains only these variants.

This would be used in `Elm.Syntax.Declaration.Destructuring` and `Elm.Syntax.Expression.FunctionImplementation`.

PR #83, opened 2020-06-27
Add destructur pattern

MartinSStewart:
Fixes https://github.com/stil4m/elm-syntax/issues/76


src/Elm/Syntax/DestructurPattern.elm
```diff
@@ -0,0 +1,210 @@
...
+  - `UnConsPattern`: `x :: xs`
+  - `ListPattern`: `[ x, y ]`
+  - `VarPattern`: `x`
+  - `NamedPattern`: `Just _`
+  - `AsPattern`: `_ as x`
+  - `ParenthesizedPattern`: `( _ )`
+
+-}
+type DestructurPattern
+    = AllPattern_
```
MartinSStewart, line 63: I chose to add underscores to all the variants so there isn't any name collisions with `Pattern`.

MartinSStewart:
I went ahead and also made it so Lambda has a nonempty number of args (since `\ -> ` is invalid)

src/Elm/Syntax/DestructurPattern.elm
stil4m: I understand. For now I think it is fine. Personally I'm not really fond of `_` I think i I'm planning to renaming all the pattern stuff. Eliminate the `Pattern` in the value constructors and 'enforce' reference by module name:
`Pattern.All` and `DestructurAll`.

src/Elm/Syntax/DestructurPattern.elm
```diff
@@ -0,0 +1,181 @@
+module Elm.Syntax.DestructurPattern exposing
```
stil4m, line 1: Is this name correct? (I'm not a native English speaker).

The Elm documentation uses 'destructure' instead of 'destructur' (https://elm-lang.org/docs/records).

MartinSStewart: You're right. That makes two times I've misspelled this word. I'll fix it now.

PR #84, opened 2020-06-28
Extension record should have nonempty number of fields

MartinSStewart:
`Elm.Syntax.TypeAnnotation.TypeAnnotation` has a `Record` variant that used to pull double duty as both a record and an extension record. It had an impossible state where it could both be an extension record and have no fields (which the Elm compiler doesn't allow).

To fix this I added a `ExtensionRecord` variant that has a nonempty number of fields.

### DIFF
--- a/src/Elm/Syntax/TypeAnnotation.elm
+++ b/src/Elm/Syntax/TypeAnnotation.elm
@@ -16,7 +16,7 @@ For example:
 
 ## Types
 
-@docs TypeAnnotation, RecordDefinition, RecordField
+@docs TypeAnnotation, RecordField
 
 
 ## Serialization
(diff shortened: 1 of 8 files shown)


src/Elm/Syntax/TypeAnnotation.elm
```diff
@@ -46,16 +47,11 @@ type TypeAnnotation
...
-    | Record RecordDefinition (Maybe (Node String))
+    | Record (List (Node RecordField))
+    | ExtensionRecord (Node String) (Node RecordField) (List (Node RecordField))
     | FunctionTypeAnnotation (Node TypeAnnotation) (Node TypeAnnotation)
 
 
-{-| A list of fields in-order of a record type annotation.
--}
-type alias RecordDefinition =
-    List (Node RecordField)
```
MartinSStewart, line 56: I removed `RecordDefinition` as it felt like unnecessary indirection when figuring out the type signature for `Record`.

PR #88, opened 2020-08-24
Fix associativity for infix operators

### LINKED ISSUE #87: Operator associativity doesn't seem to be taken into account
Taking `|>` as an example which is left-associative this expression: `1 |> 2 |> 3` should result in the following tree:

- `|>`
  - `|>`
    - 1
    - 2
  - 3

But currently it creates the following:

- `|>`
  - 1
  - `|>`
    - 2
    - 3

Here's an example:
https://ellie-app.com/9GDNxTn3ZDTa1

AttilaMihaly:
closes #87


MartinSStewart:
Looks good, thanks for the PR (and sorry about the delay)

PR #93, opened 2021-01-21
Backport fixes from the master branch to the v7 branch

### LINKED ISSUE #92: Range miscalculation when using If expressions
```elm
module TestModule exposing (..)

a =
  if cond then
    1
  else
    2
  
-- hello

{-| doc
-}
b = 3
```

In the code above, the range for `a` ends where `b` starts, meaning it will include the comment in between and the documentation for `b`.

If the if expression is removed, like here:
```elm
a =
  2
  
-- hello

{-| doc
-}
b = 3
```

then `a`'s range seems to be correct.

This looks very similar to https://github.com/stil4m/elm-syntax/issues/63 which had the same problem but for let expressions.

jfmengels:
Before releasing a new major version, I would like to publish a new v7 version with all the fixes currently present on `master`, and all the issues that have currently been reported.

This PR contains cheer-picks of everything that didn't seem to be meant to change the API of the package. The only remaining thing is to remove the breaking change introduced by fce06d96583906f84c863ce8b48bd8e6adc2e921 (EDIT: It's done), although it might be interesting to fix #92, because that sounds very related, before undoing the breaking change.

### DIFF
--- a/src/Elm/Syntax/Exposing.elm
+++ b/src/Elm/Syntax/Exposing.elm
@@ -58,7 +58,7 @@ type TopLevelExpose
 {-| Exposed Type
 -}
 type alias ExposedType =
-    { name : Node String
+    { name : String
     , open : Maybe Range
     }
 
@@ -169,7 +169,7 @@ encodeTopLevelExpose =
 encodeExposedType : ExposedType -> Value
 encodeExposedType { name, open } =
     JE.object
-        [ ( "name", Node.encode JE.string name )
+        [ ( "name", JE.string name )
         , ( "open", open |> Maybe.map Range.encode |> Maybe.withDefault JE.null )
         ]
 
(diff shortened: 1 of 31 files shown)


src/Elm/Syntax/Exposing.elm
```diff
@@ -58,7 +58,7 @@ type TopLevelExpose
 {-| Exposed Type
 -}
 type alias ExposedType =
-    { name : String
+    { name : Node String
```
MartinSStewart, line 61: Isn't this a breaking change?

jfmengels: Yes, that is the breaking change I mentioned in the description. I had a quick look at removing it, but I'm not yet familiar enough with the code and didn't succeed. I'll try again later.

jfmengels:
Alright, I was able to remove the breaking change (much easier after a good night's sleep :smile: ) :+1:

src/Elm/Syntax/Exposing.elm
MartinSStewart: Oops, my bad. Easier to read PRs after a good nights sleep as well!

PR #94, opened 2021-01-23
Fix if expression range

MartinSStewart:
This fixes https://github.com/stil4m/elm-syntax/issues/92. 

As @jfmengels mentioned, this is indeed related to an earlier `let...in` range miscalculation. I think there might be more bugs like this (though I didn't find anymore right now) as it's caused by `Node.parser` which uses the current parser position as the end position of a range rather than using the end position of the child expression. For many cases, the current parser position and the end position of the expression are the same. But in `let`, `case`, and `if` expressions, this isn't true.

### DIFF
--- a/src/Elm/Parser/Declarations.elm
+++ b/src/Elm/Parser/Declarations.elm
@@ -565,22 +565,32 @@ numberExpression =
 
 ifBlockExpression : Parser State (Node Expression)
 ifBlockExpression =
-    Node.parser
-        (ifToken
-            |> Combine.continueWith
-                (lazy
-                    (\() ->
-                        succeed IfBlock
-                            |> Combine.ignore (maybe Layout.layout)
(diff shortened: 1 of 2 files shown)


MartinSStewart:
@jfmengels It just occurred to me that maybe I should have made this PR for your backport-fixes-to-v7 branch instead of master. Let me know what you think.

jfmengels (approved):
LGTM :+1: 
I've tried it out on `elm-review` rules that had bugs around this, and they now pass.

I'll push these commits directly on the branch for https://github.com/stil4m/elm-syntax/pull/93

jfmengels:
> It just occurred to me that maybe I should have made this PR for your backport-fixes-to-v7 branch instead of master. Let me know what you think.

Keep on doing them on master. Cherry-picking them on v7 will rarely be difficult I think.

Issue #89, opened 2020-10-08
Incorrect parsing of the contents of triple-quoted strings


jfmengels:
When encountering double backslashes (`\\`) in triple quotes, the order of characters is wrong.

```elm
module A exposing (..)
_ = """\\{"""
_ = """\\{\\}"""
_ = """\\}\\{"""
_ = """\\[\\]"""
_ = """\\(\\)"""
_ = """\\a\\b"""
_ = """\\a-blablabla-\\b"""
```

For instance `\\{\\}` becomes `\\}\\{`

To play with those interactively, I made an Ellie. It's in Elm code, so there is a lot of escaping unfortunately https://ellie-app.com/bbmqddVmJzka1 (see line ~65)

PR #95, opened 2021-01-23
Fix multilines strings

MartinSStewart:
This fixes https://github.com/stil4m/elm-syntax/issues/89

### DIFF
--- a/src/Elm/Parser/Tokens.elm
+++ b/src/Elm/Parser/Tokens.elm
@@ -204,12 +204,12 @@ multiLineStringLiteral =
         helper s =
             if s.escaped then
                 escapedCharValue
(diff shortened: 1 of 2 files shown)


src/Elm/Parser/Tokens.elm
```diff
@@ -204,12 +204,12 @@ multiLineStringLiteral =
         helper s =
             if s.escaped then
                 escapedCharValue
-                    |> Core.map (\v -> Loop { s | escaped = False, parts = String.fromList [ v ] :: s.parts })
+                    |> Core.map (\v -> Loop { s | escaped = False, parts = String.fromChar v :: s.parts })
```
MartinSStewart, line 207: This isn't necessary for the bug fix but it seemed strange to do `Char to List Char to String` instead of `Char to String`

jfmengels:
This looks good, thanks @MartinSStewart :+1:

Issue #75, opened 2020-06-21
Found Operator in syntactically incorrect Elm code


jfmengels:
I found a case where we get a `Operator` element, but the code is not considered syntactically correct from Elm's point of view.

```elm
module TestModule exposing (..)

a = ( +)
```

The same applies for `a = (+ )` and `a = ( + )`.

[Interactive Ellie](https://ellie-app.com/9cqDd7R3ys6a1)

MartinSStewart:
I tried writing a test for this on master but the test correctly failed to parse.
```elm
sample40 : String
sample40 =
    """module Foo exposing (..)

x =
  (+ )

"""
```
This issue seems to have been resolved already?

PR #96, opened 2021-01-23
Add regression tests for (+ ) bug

MartinSStewart:
This fixes https://github.com/stil4m/elm-syntax/issues/75.

Or rather, that bug appears to already be fixed but I'm not sure when it happened. I've just added some regression tests.


jfmengels:
Alright. Let's just add the regression tests :+1: and consider the issue solved until someone reports it again.

Issue #47, opened 2020-05-14
Problem with non-ascii characters in variable names


woylie:
I encountered a problem with elm-syntax while using elm-analyse on the elm-visualization package.

@jfmengels pointed out that the problem is caused by the variable names `ε2` and `ρ`.

Original issue opened here: https://github.com/stil4m/elm-analyse/issues/242

> Elm-analyse returns a `FileLoadError` (`Could not load file due to: Unexpected parse error`) for the module `Zoom.Interpolation` of elm-visualization.
> 
> https://github.com/gampleman/elm-visualization/blob/2.1.1/src/Zoom/Interpolation.elm

MartinSStewart:
I had a look at this. It looks like elm-syntax checks that function names contain only alpha numeric characters. This excludes letters outside of the ascii alphabet. 

There doesn't seem to be an easy fix to this because elm-syntax needs to know if a character is uppercase or lowercase and Elm doesn't provide any API for doing that in general with Unicode characters.

MartinSStewart:
I thought of a hacky solution. We can exploit the fact that String.toLower and String.toUpper do support unicode to implement our own Char.isLower and Char.isUpper
```elm
isLower char = 
    let 
         stringChar = String.fromChar char
    in
    case (String.toUpper stringChar == stringChar, String.toLower.toLower stringChar == stringChar) of
        (False, True) -> True
        _ -> False

isUpper char = 
    let 
         stringChar = String.fromChar char
    in
    case (String.toUpper stringChar == stringChar, String.toLower.toLower stringChar == stringChar) of
        (True, False) -> True
        _ -> False
```

PR #97, opened 2021-01-23
Handle unicode in tokens

MartinSStewart:
This fixes https://github.com/stil4m/elm-syntax/issues/47

As I mention here https://github.com/stil4m/elm-syntax/issues/47#issuecomment-766186582 this solution is pretty hacky and might allow for situations where you can use some exotic unicode character in a type name or function that the Elm compiler wouldn't allow. Still, I think this is better than nothing.


jfmengels (approved):
This looks like an improvement, thanks @MartinSStewart :)

Issue #98, opened 2021-01-25
Elm.Writer generates invalid case expressions


MartinSStewart:
The following is code generated with Elm.Writer
```elm
codec : Codec MyType
codec =
    Codec.customType (\a b c value -> case value of
  VariantA ->
    a
  VariantB ->
    b
  VariantC ->
    c)
```
This is not valid Elm code for two reasons:
1. The case expression patterns aren't indented past the start of the case keyword
2. The closing parenthesis for the lambda expression is not on its own line

PR #99, opened 2021-01-25
Fix indentation in case expression

MartinSStewart:
This fixes https://github.com/stil4m/elm-syntax/issues/98

It has the side effect of adding some unnecessary new lines before and after case expressions*. To me this doesn't seem like a big issue since the Elm code will still compile and elm-format will tidy things up.

@jfmengels I'll go ahead and backport this one as it requires some changes before it will work with elm-syntax v7.

\* The reason adding a leading new line to case expressions is so the `case expr of` is on it's own line. Otherwise, it can get pushed too far to the right by the code preceding it. The trailing new line is to preventing closing parenthesis from ending up on the same line as the last case pattern (which will cause a compiler error).

### DIFF
--- a/src/Elm/Writer.elm
+++ b/src/Elm/Writer.elm
@@ -518,8 +518,10 @@ writeExpression (Node range inner) =
                             ]
             in
             breaked
-                [ spaced [ string "case", writeExpression caseBlock.expression, string "of" ]
+                [ string ""
+                , spaced [ string "case", writeExpression caseBlock.expression, string "of" ]
                 , breaked (List.map writeCaseBranch (caseBlock.firstCase :: caseBlock.restOfCases))
+                , string ""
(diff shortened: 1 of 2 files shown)


jfmengels (approved):
Alright. The blank lines where unnecessary is annoying (not everyone runs or can run `elm-format` on generated code), but it's definitely better than before.

It would be good to find a solution to this at some point though, maybe through passing an indentation level as an argument to `writeExpression` or one of it's helper functions?

MartinSStewart:
I think the best approach might be to copy the implementation in https://package.elm-lang.org/packages/the-sett/elm-syntax-dsl/latest/. It claims to closely follow elm-format. If it can do that then it's probably much nicer than Elm.Writer currently is.

PR #103, opened 2021-01-30
Speed up how comments are matched with declarations

jfmengels:
Instead of filtering through all the comments that matched then taking the head, we go through the list and stop once we find one that matches.

This relies on comments already being ordered (otherwise it still works, but it won't be as fast), which was added in #102.

I don't have any benchmarks for this though. We don't need this for the 7.1.4 patch release.

### DIFF
--- a/src/Elm/Processing/Documentation.elm
+++ b/src/Elm/Processing/Documentation.elm
@@ -26,11 +26,7 @@ postProcess file =
 
 onType : Node Type -> File -> File
 onType (Node r customType) file =
-    let
-        docs =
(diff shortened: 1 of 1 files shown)


MartinSStewart:
I'm not sure what you mean by ordering matters. It seems like stopping early will just make this take half as much time on average? Eitherway, it looks good so I merged it.

jfmengels:
I meant that with the comments being in the same order as the declarations (and with the found one being removed every time), we the finding operation in itself will be faster than if they were in the reverse order.
So this is an improvement probably, but with the right order even more.

PR #105, opened 2021-01-31
Add changelog for 7.2.0

jfmengels:


MartinSStewart:
Do you know why the AppVeyor builds are failing in some PRs? It seems to happen only to PRs that are not merging into master.

jfmengels:
There has been a fix for that on `master`, but simply cherry-picking it onto the v7 branch doesn't work.

MartinSStewart:
Strange. I'm guessing it's a problem with the CI and not with the PR though?

jfmengels:
Yes, all my PRs on the v7 have been met with this issue. I'll try again later, but I don't feel like this is blocking us. Also, it looks like Travis does the exact same checks as appveyor, so I'm really unsure as to why we have the two build setups.

PR #116, opened 2021-03-25
Fix parsing error when there are no parens around `=` in a type alias declaration

jfmengels:
There was an issue with parsing type alias declarations when there was no spacing on either side of the `=` sign.

```elm
-- Succeed to parse
type alias A = ()
-- Fail to parse
type alias A= ()
type alias A =()
```

Example of this happening in an Ellie: https://ellie-app.com/cJJBxZd4gkga1

This is also the case in v8, so we'll need to backport this to `master`.

### DIFF
--- a/src/Elm/Parser/Declarations.elm
+++ b/src/Elm/Parser/Declarations.elm
@@ -516,7 +516,7 @@ letDestructuringDeclarationWithPattern p =
             succeed (LetDestructuring p)
                 |> Combine.ignore (maybe Layout.layout)
                 |> Combine.ignore (string "=")
-                |> Combine.ignore Layout.layout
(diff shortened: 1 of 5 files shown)


MartinSStewart:
I looked at all the places `"="` shows up and found that it's also incorrect for `LetDestructuring`. I went ahead and fixed that. It looks like it's also incorrect for `destructuringDeclaration` in `Declaractions.elm` but I think this can be left alone because destructuring isn't possible for top level declarations so really that whole function should be removed at some point.

jfmengels:
Nice :+1: 
Your changes look good to me. Are you okay with merging this then?

MartinSStewart:
Sounds good to me

MartinSStewart (approved):
Looks good!

PR #118, opened 2021-03-25
Indent Types correctly

ceddlyburge:
This change is already in the main branch.

Without it Types are output like this

```
Type Cedd
= Cedd String
```

And with it they look like this

```
Type Cedd
    = Cedd String
```


MartinSStewart:
I went ahead and fixed the formatting and a failing test (the test is incorrect). Thanks for the fix!

ceddlyburge:
Thanks Martin :)

PR #121, opened 2021-03-29
 Replace isLower and isUpper workaround with miniBill/elm-unicode

MartinSStewart:
Unicode aware `isLower` and `isUpper` functions were implemented using an ugly workaround involving `Char.toLower` and `Char.toUpper`. `elm-unicode` is published so we can use that instead now.


MartinSStewart:
I made the same mistake twice...

Issue #124, opened 2021-04-07
CharLiteral doesn’t handle escapes correctly


NiavlysB:
In [`writeExpression`](https://github.com/stil4m/elm-syntax/blob/609c0ee58c7ebf65628cea49bfd8044c0aff9fe5/src/Elm/Writer.elm#L496), the CharLiteral case doesn’t check if the character is an escape (backslash `\\` for instance).

        CharLiteral c ->
            string ("'" ++ String.fromList [ c ] ++ "'")

This seems to produce a syntax error in the generated Elm code, because the single-quote delimiters aren’t closed:

    '\'

It should probably also escape an actual tab to a `'\t'`, a `'→'` to a `'\u{2192}'`, etc.

MartinSStewart:
Hello, I'll fix this over the weekend. That said, I recommend using elm-syntax-dsl instead of the built-in Elm.Writer. elm-syntax-dsl is more stable and generates better formatted code.

NiavlysB:
I am actually using elm-syntax-dsl! But looking up the code I figured it was relying at the lowest lovel (char literal) on Elm.Writer. Maybe I’m wrong?

MartinSStewart:
Maybe, I'll take a closer look when I go to fix this.

NiavlysB:
Thanks!

MartinSStewart:
This should be fixed now once a patch version of elm-syntax is released. Also I checked and I don't believe this bug affected elm-syntax-dsl as none of the modules in elm-syntax-dsl import Elm.Writer. It probably has its own implementation.

NiavlysB:
Oh ok, I’ll have a closer look at elm-syntax-dsl then! Thanks!

PR #125, opened 2021-04-10
Fix CharLiteral not escaping

MartinSStewart:
Fixes https://github.com/stil4m/elm-syntax/issues/124.

The issue mentioned that `→` should be converted to `\u{2192}` in the generated code. I don't think this should be done. Whether a unicode code point should be represented directly or with the \u{} notation is a choice the programmer (or elm-format) can make. As far as I can tell, both should work. For that reason, Elm.Writer will always write the actual unicode character as it's easier to do.

### DIFF
--- a/src/Elm/Writer.elm
+++ b/src/Elm/Writer.elm
@@ -490,7 +490,16 @@ writeExpression (Node range inner) =
             string ("\"" ++ s ++ "\"")
 
         CharLiteral c ->
-            string ("'" ++ String.fromList [ c ] ++ "'")
+            let
+                escape : String
+                escape =
+                    if c == '\t' || c == '\'' || c == '\\' then
+                        "\\"
+
+                    else
+                        ""
+            in
+            string ("'" ++ escape ++ String.fromList [ c ] ++ "'")
 
         TupledExpression t ->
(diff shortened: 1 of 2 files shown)


src/Elm/Writer.elm
```diff
@@ -490,7 +490,16 @@ writeExpression (Node range inner) =
...
+            let
+                escape : String
+                escape =
+                    if c == '\t' || c == '\'' || c == '\\' then
+                        "\\"
+
+                    else
+                        ""
+            in
+            string ("'" ++ escape ++ String.fromList [ c ] ++ "'")
```
jfmengels, line 502: Couldn't we use String.fromChar instead?

jfmengels (approved):
LGTM

src/Elm/Writer.elm
MartinSStewart: Good point. I checked where String.fromList gets used and noticed that CharPattern isn't escaped either so I fixed that now.

NiavlysB:
> The issue mentioned that → should be converted to \u{2192} in the generated code. I don't think this should be done.

Totally, I’m not sure why I added this point in the issue. I like it better with the actual symbols too :)

PR #131, opened 2021-07-14
Fix `Expression.RecordAccessFunction` parser

SiriusStarr:
Currently, `Expression.RecordAccessFunction` includes a leading `.` in its parsed string, i.e. the function in

```elm
List.map .name xs
```

parses as  `Expression.RecordAccessFunction ".name"`.  This is inconsistent with `Expression.RecordAccess`, where

```elm
x.name
```

parses as `Expression.RecordAccess _ (Node _ "name")` and, more importantly, inconsistent with `Writer.writeExpression`, which prepends a `.`:

```elm
        RecordAccessFunction s ->
            join [ string ".", string s ]
```

This currently leads to invalid Elm being written by `Writer.writeExpression` if it is used to write a parsed `RecordAccessFunction`, as parsing and then writing e.g.

```elm
List.map .name xs
```

would produce

```elm
List.map ..name xs
```

### DIFF
--- a/src/Elm/Parser/Declarations.elm
+++ b/src/Elm/Parser/Declarations.elm
@@ -663,7 +663,7 @@ referenceExpression =
 
 recordAccessFunctionExpression : Parser State (Node Expression)
 recordAccessFunctionExpression =
-    Combine.map ((++) "." >> RecordAccessFunction)
+    Combine.map RecordAccessFunction
         (string "."
             |> Combine.continueWith functionName
         )
--- a/src/Elm/Parser/Declarations.elm
+++ b/src/Elm/Parser/Declarations.elm
@@ -663,7 +663,7 @@ referenceExpression =
 
 recordAccessFunctionExpression : Parser State (Node Expression)
 recordAccessFunctionExpression =
-    Combine.map RecordAccessFunction
+    Combine.map ((++) "." >> RecordAccessFunction)
         (string "."
             |> Combine.continueWith functionName
         )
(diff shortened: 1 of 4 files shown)


jfmengels:
I agree that this should be changed. Thanks @SiriusStarr !

That said, I think this should probably not be backported to the v7 branch (master is our "V8" branch, it has plenty of breaking changes), as it would be a surprising change and some tools might stop working as expected. What do you think @MartinSStewart ?

MartinSStewart:
I agree, I think this is too risky to include in the current version. But it should definitely be included in the next major release.

I can take a closer look at this PR later today. Thanks @SiriusStarr !

SiriusStarr:
Understandable; I just used `v7` since the tests are failing for `master` (looks like lambda arguments got switched to be non-empty but haven't been in the tests yet) so I couldn't be sure I wasn't breaking anything.  :rofl: 

I can tweak it to fix `Writer.writeExpression` for now and the parser change can be put off till v8?

MartinSStewart:
> I can tweak it to fix Writer.writeExpression for now and the parser change can be put off till v8?
Sounds good!

MartinSStewart:
@SiriusStarr Thanks for pointing out that the tests weren't compiling on master. I've fixed that now.

SiriusStarr:
Alright, I've reverted the parser changes and instead made the writer check for a leading `.` and not double append if it finds one, so it should write valid code with both:

```
Current behavior:
Expression.RecordAccessFunction "name"
```

and

```
New behavior:
Expression.RecordAccessFunction ".name"
```

now.

MartinSStewart:
Cool, that's a smart way to avoid a breaking change. Thanks again for the PR!

MartinSStewart:
There's another bug I'm going to fix over the weekend and then I'll release a new patch version.

PR #132, opened 2021-07-17
Fix let function declaration range

MartinSStewart:
Previously, function declarations in a let expression would include all the trailing whitespace and even the `in` keyword in their range. I've fixed this so now the let function declaration's range stops at the end of their respective expressions.


jfmengels (approved):
LGTM :)
Seems like the library has been bitten quite often by tests that ignore the range now...

PR #133, opened 2021-07-17
Add elm review

MartinSStewart:
I've added elm-review to this repo and ran these rules
```elm
config : List Rule
config =
    [ NoUnused.Dependencies.rule
    , NoUnused.Variables.rule
    , NoUnused.Modules.rule
    ]
```
I decided to not add anything else for now. I don't want the modules to changes too drastically and make it even harder to copy fixes from v7 to master.


jfmengels (approved):
Looks good.

We should replace the testing suite soon though with GitHub Actions. It becomes a bit hard to know whether the test suite passed or not now.

PR #138, opened 2021-08-03
Build / dev environment improvements

jfmengels:
Plenty of small things. It's readable commit by commit.
The PR is intimidating because of all the big files like yarn.lock, but not too much happened.

I took a lot of things from some of my other packages (like `elm-review-unused`)

I tried removing appveyor, but I think that requires accessing the settings of the project, which I don't have access to.


MartinSStewart:
You removed appveyor right? If so, why does the PR say it's failing?

jfmengels:
I think it's an integration that was enabled through the repo's settings, and then additionally configured through a configuration file. I think we'll need @stil4m's help to remove that, as we don't have access to the Settings of the repo.

MartinSStewart:
Does the CI publish step only try to publish commits on v7? Or maybe it doesn't matter since the user needs to add a tag so there no risk of accidentally publishing early?

jfmengels:
The publish step is only done when the version of the package (in `elm.json`) gets changed to a value that has not been published yet (and tests pass), no need to add a tag (the action will do that automatically). The version on `master` is still set to an old version. But we should avoid setting it to v8 until we're ready!

I'm not entirely sure whether this will work for v7, or if it will only do it will only be for master. Likely it will only work for master. If you find this scary, I can disable this until v8 is our main target. And/or we could make v7 our master branch and rename `master` to `v8`.

MartinSStewart:
Yeah, I find that a bit scary. I don't think it's worth the small amount of effort it will save.

jfmengels:
I *really* like it, it makes publishing a breeze. But I'll remove it for now, so that we don't publish v8 before its time 👍

MartinSStewart:
Yeah, I guess, accidentally publishing a patch is not big deal (maybe a little awkward). Accidentally publishing a major release is a *big* issue. I agree, lets wait until we're on v8

MartinSStewart:
Anyway, I'm satisfied then. Go ahead and merge 👍

PR #139, opened 2021-08-04
Backport setup improvements

jfmengels:
Backporting all the changes from https://github.com/stil4m/elm-syntax/pull/138

I for no disabled elm-review because none of the fixed applied on master have been backported here.


MartinSStewart:
Sounds good. And the CI publish will be enabled on this branch so all one should do is bump the elm.json version?

jfmengels:
I've disabled it as well on this branch. From the documentation on https://github.com/dillonkearns/elm-publish-action it looks it only works for the main branch, so we'd have to make v7 the main branch (which is maybe not a bad idea)

PR #140, opened 2021-08-04
Add elm-review config

jfmengels:
- Add elm-review configuration with some basic rules, and enforces them
- Add elm-review to the test suite


jfmengels:
Yeah. I tried doing about the same things so we wouldn't have to do a lot of work backporting (which I plan on taking on anyway)

PR #141, opened 2021-08-04
Speed up adding documentation

jfmengels:
I noticed that `Elm.Processing.Documentation` we were doing a visit of the entire AST, due to the implementation of the `Inspector`. `visitors` were not being set (marked as `Continue`), but the inspector wasn't being very smart and visited every node anyway.

A lot of those visits were entirely unnecessary, so I removed those. Then, I noticed things could be very much simplified, so I ended up rewriting everything bit by bit.

When I got to that point, I noticed that we were looping over declarations and comments plenty of times, even though we know they're both ordered by range, so I made a more optimized version where we try to abort earlier and avoid re-checking comments.

Next step (just writing it so I don't forget it): Try to do the pass about re-arranging operator trees in the same pass as the one where we add documentation.

I'll take another review myself at this PR later, but feel free to review or ask questions in the meantime. I guess we should do a sanity check that this results in the same AST as before.


MartinSStewart:
Should this PR be merging these changes into master, instead of review-v7?

jfmengels:
It was based on some of the changes done in `review-v7`, so I needed to get that merged in first. Now that that one is merged, it's automatically pointing to master again :)

MartinSStewart:
Ah, that makes sense. Go ahead and merge this one as well then 👍

PR #144, opened 2021-08-05
Speed up adding documentation (Backport of #141)

jfmengels:
Simple backport of what was needed to backport #141, with some additional changes to support port declarations with comments.

I have to admit I don't understand some parts of how the documentation is attached (same on master and v8), because when documentation for a declaration (not including port) is found, we don't update the range, we take the existing one. I assume the existing one wasn't updated to include the documentation range before this step, yet after the step it's correct, even though we don't update the range.

I don't get it but tests pass.


jfmengels:
Seems like this was indeed a bug. It was fixed in #145, which I'll backport here later after it gets merged.

PR #155, opened 2021-10-13
Fix erroneous greedy lambda range parsing

### LINKED ISSUE #136: Lambda expression range is miscalculated when not parenthesized and overlaps next expression by one character
In

```elm
module A exposing (..)


f = \i -> i
g x =
    x
```
the lambda expression parses to the node `Node { end = { column = 1, row = 5 }, start = { column = 5, row = 4 } } (LambdaExpression...`, which has it ending on row 5, overlapping the beginning of `g x =` by one character

When parenthesized, this does not occur, e.g.

```elm
module A exposing (..)


f = (\i -> i)
g x =
    x
```
parses to `Node { end = { column = 14, row = 4 }, start = { column = 5, row = 4 } } (ParenthesizedExpression (Node { end = { column = 13, row = 4 }, start = { column = 6, row = 4 } } (LambdaExpression`

To generate this behavior in the REPL (since multiline strings don't copypaste nicely:

```elm
parse "module A exposing (..)\n\n\nf = \\i -> i\ng x =\n    x\n"

parse "module A exposing (..)\n\n\nf = (\\i -> i)\ng x =\n    x\n"
```

And here's a character diff of the outputs to illustrate:
RJqDijE

SiriusStarr:
Fixes #136.  Looks like this got missed in 9d7a87a9b21f77b6ab8d76e188fa1a1c7089e4a9 which fixed it for `let` but not lambdas.  This just applies the same fix and adds a test for it.

These changes should probably be backported to `v7`, since they are currently causing `elm-review` rules to generate invalid code, e.g SiriusStarr/elm-review-no-unsorted#3.

### DIFF
--- a/src/Elm/Parser/Declarations.elm
+++ b/src/Elm/Parser/Declarations.elm
@@ -398,12 +398,19 @@ lambdaExpression : Parser State (Node Expression)
 lambdaExpression =
     lazy
         (\() ->
-            succeed (\args expr -> Lambda args expr |> LambdaExpression)
-                |> Combine.ignore (string "\\")
-                |> Combine.ignore (maybe Layout.layout)
-                |> Combine.andMap (sepBy1 (maybe Layout.layout) functionArgument)
-                |> Combine.andMap (Layout.maybeAroundBothSides (string "->") |> Combine.continueWith expression)
-                |> Node.parser
+            Ranges.withCurrentPoint
+                (\current ->
+                    succeed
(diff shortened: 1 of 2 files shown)


jfmengels (approved):
LGTM. @MartinSStewart would you mind having a look?
I'll take care of backporting and releasing if you're good with this change.

MartinSStewart:
I'll take a look tomorrow 👍

PR #159, opened 2022-02-07
fix(writer): adds parenthesis around contructor declaration arguments

### LINKED ISSUE #158: writeDeclaration doesn't print parenthesis around type constructor argument
The following test case will fail.

```elm
            , test "write type declaration (with constructor arguments)" <|
                \() ->
                    let
                        listType =
                            Elm.Syntax.TypeAnnotation.Typed (Node emptyRange ( [], "List" ))
                                [ Node emptyRange <|
                                    Elm.Syntax.TypeAnnotation.Typed (Node emptyRange ( [], "String" )) []
                                ]
                    in
                    (Node emptyRange <|
                        CustomTypeDeclaration
                            (Type
                                Nothing
                                (Node emptyRange "Sample")
                                []
                                [ Node emptyRange <| ValueConstructor (Node emptyRange "Foo") [ Node emptyRange listType ]
                                , Node emptyRange <| ValueConstructor (Node emptyRange "Bar") []
                                ]
                            )
                    )
                        |> Writer.writeDeclaration
                        |> Writer.write
                        |> Expect.equal
                            ("type Sample \n"
                                ++ "    =Foo (List String) |Bar "
                            )
```

### Expected outcome

Expected it to return the constructor with parenthesis around the argument
`=Foo (List String) |Bar`

### Observed outcome

The test fails as the result from writer was `=Foo List String|Bar`. It doesn't write the parenthesis around the argument which generates broken code.

```
✗ write type declaration (with constructor arguments)

    "type Sample \n    =Foo List String|Bar "
    ╷
    │ Expect.equal
    ╵
    "type Sample \n    =Foo (List String) |Bar "
```

phenax:
Fixes #158

### DIFF
--- a/src/Elm/Writer.elm
+++ b/src/Elm/Writer.elm
@@ -280,7 +280,7 @@ writeValueConstructor : ValueConstructor -> Writer
 writeValueConstructor { name, arguments } =
     spaced
         [ string <| Node.value name
-        , spaced (List.map writeTypeAnnotation arguments)
+        , spaced (List.map (writeTypeAnnotation >> parensIfContainsSpaces) arguments)
         ]
 
 
--- a/src/Elm/Writer.elm
+++ b/src/Elm/Writer.elm
@@ -280,7 +280,7 @@ writeValueConstructor : ValueConstructor -> Writer
 writeValueConstructor { name, arguments } =
     spaced
         [ string <| Node.value name
-        , spaced (List.map (writeTypeAnnotation >> parensIfContainsSpaces) arguments)
+        , spaced (List.map (wrapInSurroundingParentheses >> writeTypeAnnotation) arguments)
         ]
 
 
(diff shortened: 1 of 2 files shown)


jfmengels:
Yes, I had the same concern. I think that there could be a better check/heuristic than `parensIfContainsSpaces` used here. I don't know if this should be blocking the PR, but I think that if you're up for it @phenax (and thank you for the PR and issue btw :pray: ), then maybe that could be improved.

To be more explicit, the idea I had is something like:
```elm
wrapInSurroundingParentheses : Node TypeAnnotation -> Node TypeAnnotation
wrapInSurroundingParentheses node =
  case Node.value node of
    TypeAnnotation.GenericType -> node
    TypeAnnotation.Typed _ typeParameters ->
        case typeParameters of
          [] -> node
          _ -> Node {- an empty range? -}  (TypeAnnotation.Tupled [ node ])
    -- ...
```
and then doing
```elm
List.map (wrapInSurroundingParentheses >> writeTypeAnnotation) arguments
```

phenax:
That sounds like a much better approach. Will work on it this weekend.

Unrelated to this, can you tell me why we use Tupled for it? Since semantically, a tuple is different from a type formatted with parens. Should the formatting instead be based on some context that defines whether or not a printed type should be wrapped in parens?

jfmengels:
Thank you @phenax for working on this :)

Just FYI, we do believe that other libraries like `the-sett/elm-syntax-dsl` do a better job at the writing part, so we are considering removing this feature in the next major version (which we're not working on actively at the moment, so you have time to migrate or to give feedback on this :smile: ).

jfmengels:
@phenax This is now published in v7.2.9

PR #176, opened 2023-06-23
Remove unnecessary lazy calls

jfmengels:
There were a large number of unnecessary `lazy` calls. Hopefully this will make parsers slightly faster. If not, then they'll at least be a bit cleaner.

EDIT: I forgot we have like 2 main branches on this project. I'll backport this to v8 afterwards.


MartinSStewart (approved):
Looks goods to me!

PR #177, opened 2023-06-23
Update elm-review configuration

jfmengels:
Updates the elm-review configuration to the most recent versions of `elm-review` and the `package` starter template, and fixes a bunch of reported errors.


MartinSStewart (approved):
Also looks good!

PR #179, opened 2023-06-24
Make sure file comments are sorted

jfmengels:
This sorts comments from closest to furthest from the start of the file.
Previously, the order was somewhat random, as can be seen in [this temporary commit](https://github.com/stil4m/elm-syntax/commit/fe59cac6aa26051483502d861da2b9dcca38d078#diff-5695ccd7e19fe3749727cf2a0f2a627f534609a7ce87f8682997c612096b1923R96-R102)

We already do this sorting in `elm-review`, but it feels off to have to do it there. 
This will make it more consistent as well, since declarations and imports are in "normal" order but comments weren't.

Can be reviewed commit by commit.


MartinSStewart (approved):
Good idea!

PR #184, opened 2023-06-25
Simplify Interface.fileToDefinitions

jfmengels:
I have to admit I am very confused as to what the function does. I *think* almost everything about this function (or even where it's end up being used) is unnecessary.

Also, this section of the project doesn't have tests :/

Maybe we should hold off on this change. Let me know what you think.

### DIFF
--- a/src/Elm/Interface.elm
+++ b/src/Elm/Interface.elm
@@ -201,9 +201,12 @@ fileToDefinitions file =
                 )
                 file.declarations
     in
+    -- I really don't understand what this part of the function does.
     allDeclarations
         |> List.map
             (\( name, _ ) ->
+                -- AFAIK, it's not possible to have two declarations with the same name
+                -- in the same file, so this seems pointless.
                 allDeclarations
                     |> List.filter (\( otherDeclarationName, _ ) -> otherDeclarationName == name)
             )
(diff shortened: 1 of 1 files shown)


src/Elm/Interface.elm
```diff
@@ -202,17 +201,16 @@ fileToDefinitions file =
...
-        |> List.map Tuple.first
-        |> List.Extra.unique
         |> List.map
-            (\x ->
-                ( x
-                , allDeclarations
-                    |> List.filter (Tuple.first >> (==) x)
-                )
+            (\( name, _ ) ->
+                -- AFAIK, it's not possible to have two declarations with the same name
```
MartinSStewart, line 208: I mean, it is possible, it's just not valid Elm code. That said, it doesn't seem like it should be a syntax error so maybe elm-syntax should allow multiple declarations to appear?

jfmengels: You're right. I'm not sure what to would be best here. In practice I don't think it matters at all. Shall we just close this PR? This is about simplifications/performance anyway, so not something super important.

MartinSStewart: Sure we can revisit this if it ever become a problem in practice

Issue #108, opened 2021-02-02
Store raw string for numbers in the AST


jfmengels:
Currently, if we try to parse the following code

```elm
a = 100000000000000000000
```

we'll get an AST with the following node
```elm
Integer 9.999999999999999e+29
```

It's not much, but there is a precision loss. If we were to write that number using the writer, we would not get the same output as the input.

My thinking is that we could keep the original "text" in the same node, like

```elm
Integer 9.999999999999999e+29 "100000000000000000000"
```

and then the writer could re-use that original text from the AST. `elm-review` rules could also use this information for other reasons, though I can't think of any.

Downside: It's possible to have different values represented by `a` and `b` in `Integer a b` (though that is also true currently in a way...), but now it's even easier because you can do `Integer 1 "2"`, or even `Integer 1 ""`.

I think the `Floatable` variant should use the same approach.

I'm thinking we could remove the `Hex` variant in favor of `Integer Int String`. I think in most cases (at least in `elm-review` rules), we don't care how an integer is written but we care about the value more. If we care about integers written in hex, it's reasonably easy to check whether the string starts with `0x`.

If Elm ever adds support for writing numbers like `10_000`, then we will need to update the parser but we won't need to change the `Expression` type.

(This was inspired by issue https://github.com/avh4/elm-format/issues/690, where the problem seems to appear for smaller numbers than this one.)

MartinSStewart:
Will a user want to do math with the Integer or Floatable value? If it contains a string instead of an Int or Float then we've left it up to them to parse the string and they might forget to handle edge cases like hex values. Maybe we could instead have an opaque type that contains the number and helper functions like `isHex: IntegerValue -> Bool`, `toInt : IntegerValue -> Maybe Int`, `toString : IntegerValue -> String`, etc?

jfmengels:
I don't think people will do math on the expressions no, but I think people will potentially want to pattern match on the numbers `Integer 1 _ ->`, which is not possible with an opaque type.

PR #189, opened 2023-08-16
Add information to Expression about which quotes a Literal is made of

### LINKED ISSUE #57: Add the information to String literal of how it was created
It would be nice to know whether a String literal was created using `"x"` or using `"""x"""`.

I don't think we have the information at this point, and if we want to pin-point a part of the string based on the range from the expression node, then we need to account for the `"""`.

Possible ways:

1. Add the information as an additional argument
```elm
type Expression =
  = ...
  | Literal StringLiteralType String

-- Let's think of a better name for this one
type StringLiteralType
  = SingleQuote
  | TripleQuote
```

Then, if we want to select a sub-string, we need to offset the string by one for `SingleQuote` and by 3 for `TripleQuote`.

2. Add a new variant to expression:
```elm
type Expression =
  = ...
  | Literal String
  | MultilineStringLiteral (List (Node String))
```

I am thinking it could be interesting to have a List of lines, each with their own range. That would make a lot of things quite easier. If you do need the concatenated string, then that is relatively easy to do.

I know I have struggled with finding sub-strings in multiline comments (maybe that could also be split into different types like this?), and I hope that this can make things a bit easier (especially if we have the range for each line).

I think this option is better, but the thing that makes it less nice is that if you are looking at strings, you need to handle both cases.

jfmengels:
Fixes #57


MartinSStewart:
If we have a StringLiteralType, should we also have a IntegerType instead of having Integer and Hex expressions?

jfmengels:
Possibly yes. There is a separate issue in https://github.com/stil4m/elm-syntax/issues/108 for that one though, so we can brainstorm that one over there.

Btw, I'm open to alternative strategies for providing the string literal type, though I think this one is pretty good :slightly_smiling_face:

MartinSStewart (approved):

MartinSStewart:
Yeah, I prefer this approach over having two different variants in string literals in Expression

Issue #174, opened 2023-06-20
`n-1` gets misparsed


miniBill:
```elm
n-1
```
gets misparsed as `n (-1)` instead of `n - 1`

jfmengels:
Fixed in #207

PR #207, opened 2023-09-03
Re-implement expression parser using a pratt parser

### LINKED ISSUE #199: Reimplement the parser using pratt parsers
In the `Expression` type, there is the `Operator String` variant which users have to handle but will in practice never appear. This is used temporarily before we post-process the parser and re-balance the AST with operators being replaced with the appropriate `OperationApplication` AST Node.

Before re-balancing, we will end up with `Application [ left, Operator "+", right ]` which will get turned into `OperatorApplication "+" infixDirection left right`.

Since users of the package will never encounter the `Operator` variant, it's better to remove it. To do so, we either need to come up with a different trick to keep the operation somewhere before we transform it, or we parse operations the right way directly, which is the option I'd like to suggest.

There is a special kind of parser (or rather, a parser technique) that makes sure that operations are balanced the correct way directly. @janiczek explains how it works in [this article](https://martin.janiczek.cz/2023/07/03/demystifying-pratt-parsers.html), and there is already a package for it: [dmy/elm-pratt-parser](https://dark.elm.dmy.fr/packages/dmy/elm-pratt-parser/latest/).

I would like to try this out, if we are successful, we can then get rid of the post-processing of the AST, which I can only imagine will make the parser much faster. The other usage of post-processing is for attaching `{-| ... -}` comments to the module or to the different declarations, which I believe we should also be able to do without post-processing.

---

I have already started a branch for this called `pratt-parser-v8`, which I made on top of the #188 branch when I realized this was needed to get rid of the `Operator` variant.

I am now realizing it is probably better to do this work **before** v8, in a v7 release.

The main reason for that is I am pretty new to writing parsers, and I find it really hard to adapt the parser for this package (which was made before `elm/parser` and then adapted using a wrapper), so... I'm thinking of re-writing the parser almost from scratch using pratt parsers and inspiration of the existing code. But I don't have confidence I can do this correctly the first time around, so I will probably want to compare the results for the new parser with the existing parser. And that is easy to do in the v7 branch when the ASTs are the same, but much harder to do when the.

I think this will be a good situation to add more tests too, especially for cases where the code should not parse, which is useful for `elm-review` for instance. We could also use the occasion to write nice errors for parsing problems.
Finishing #185 first will likely be a good addition, to make sure that we don't mess up the ranges.

I will try to start my `pratt-parser-v8` from scratch on top of v7. From my current attempt, I know I could really use some help though :sweat_smile:

jfmengels:
Fixes #199.

Reimplements the expression using a Pratt Parser (see explanation https://martin.janiczek.cz/2023/07/03/demystifying-pratt-parsers.html).

This is not completely done, and I imagine the remaining parts will be the trickiest to do. But I'm really happy about the progress (I've been trying for a few weeks but couldn't get it to work until this morning!).

I'm off for a week+ of vacation, so I'll have to pause work on this. If some people have insight about how to do this better, I'd love to hear about. If some want to continue the work, I'd love that too :smile:


jfmengels:
If you go back one commit from the current state of the branch, i.e this commit https://github.com/stil4m/elm-syntax/pull/207/commits/9126cda3ece73d00512ed46220896bf07bd2b6a5, there is a `Test.only` on a test around parsing case expressions that doesn't work. IIRC It's because in the following case expression, the `-` symbols are considered to be the `-` by the Pratt parser.

```elm
case f of
   True -> 1
   False -> 2
```

I could be wrong about this, but that's what I had figured at some point. I have not been able to find a way to fix it yet.

src/Pratt.elm
```diff
@@ -0,0 +1,445 @@
...
+
+-- EXPRESSION PARSER CONFIGURATION
+
+
+{-| An opaque type based on
+[`Pratt.Advanced.Config`](Pratt.Advanced#Config) holding the parser
+configuration.
+-}
+type alias Config state expr =
+    Advanced.Config state expr
```
jiegillet, line 54: Since the main (only?) difference between `Pratt` and `Pratt.advanced` was the type, it doesn't really make sense to separate them anymore does it?
What would you think of merging the two? Do we keep the comments?
I know we will add the license information later on.

jfmengels: Yes, we can merge them, good idea

src/Elm/Parser/Expression.elm
```diff
@@ -1,45 +1,220 @@
...
+            ]
+        , andThenOneOf =
+            [ -- Elm Basics sets the precedence of <| and |> to 0, but Pratt.expression starts at 0 so it needs to be larger
+              infixRight 1 "<|"
+            , infixLeft 1 "|>"
+            , infixRight 2 "||"
+            , infixRight 3 "&&"
+
+            -- TODO Report a syntax error when encountering multiple of the comparison operators
+            -- `a < b < c` is not valid Elm syntax
```
jiegillet, line 81: I'm on the fence about this. `a < b < c` seems like something that should parse correctly but be rejected at the next step by the compiler, a little like if you tried to define two functions with the same name.
I could be wrong though, it's right at the line between syntax and AST verification, and the Elm compiler doesn't really separate the steps (I haven't looked at the source).

jfmengels: You're right that (as far as I know) it's done at a further step.

But I would be in favor of supporting it (if it's not too perf-draining). The point is that this syntax will always be incorrect (regardless of what kind of error the compiler reports), because it would be useful for tools using this library to know when they look (elm-review, ...) at or generate (elm-codegeni think, elm-review, ...) incorrect code. Right now you could create an autofix in elm-review which would be incorrect, whereas syntax errors are prevented by elm-review.

Similarly, I think we could check when `|>` and `<|` are used together like `a |> b <| c`.

But I'm happy for this to be a stretch goal, done even after releasing the pratt-pqrser change. And if it impacts performance or complexity too much, then I'm considering not doing it.

review/src/OrderedRanges.elm
```diff
@@ -18,6 +18,24 @@ rule =
...
 expressionVisitor node =
     case Node.value node of
+        Expression.Application [ Node _ (Expression.FunctionOrValue [] "Application"), Node _ (Expression.ListExpr ((Node firstItemRange _) :: _)) ] ->
+            let
+                { start, end } =
+                    Node.range node
+            in
+            if start.row == end.row then
+                [ Rule.errorWithFix
+                    { message = "Split Application on multiple likes"
```
jiegillet, line 28: Nice rule 💯 
```suggestion
                    { message = "Split Application on multiple lines"
```

jfmengels: Good catch :+1: 
And I had good inspiration :wink: 
I wonder what other node we can apply this to, I don't remember what I usually did manually.

jiegillet:
I usually spread out operators too, and types that are recordsSent from my iPhoneOn Jul 8, 2024, at 16:11, Jeroen Engels ***@***.***> wrote:﻿
@jfmengels commented on this pull request.



In review/src/OrderedRanges.elm:
> @@ -18,6 +18,24 @@ rule =
 expressionVisitor : Node Expression -> List (Error {})
 expressionVisitor node =
     case Node.value node of
+        Expression.Application [ Node _ (Expression.FunctionOrValue [] "Application"), Node _ (Expression.ListExpr ((Node firstItemRange _) :: _)) ] ->
+            let
+                { start, end } =
+                    Node.range node
+            in
+            if start.row == end.row then
+                [ Rule.errorWithFix
+                    { message = "Split Application on multiple likes"

Good catch 👍
And I had good inspiration 😉
I wonder what other node we can apply this to, I don't remember what I usually did manually.

jiegillet:
BTW: fixes #174

PR #210, opened 2024-04-13
Fix case..of parser by fixing indentation stack

Janiczek:
Only the first commit 3ae7c81 deals with the bug mentioned [here](https://github.com/stil4m/elm-syntax/pull/207#issuecomment-2053762106)

The rest are just some general cleanup. Feel free to cherry-pick whatever you want!

### DIFF
--- a/src/Elm/Parser/State.elm
+++ b/src/Elm/Parser/State.elm
@@ -35,7 +35,7 @@ expectedColumn (State { indents }) =
 
 pushIndent : Int -> State -> State
 pushIndent col (State s) =
-    State { s | indents = (col + 1) :: s.indents }
+    State { s | indents = (col - 1) :: s.indents }
 
 
 popIndent : State -> State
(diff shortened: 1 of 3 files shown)


src/Elm/Parser/State.elm
```diff
@@ -35,7 +35,7 @@ expectedColumn (State { indents }) =
 
 pushIndent : Int -> State -> State
 pushIndent col (State s) =
-    State { s | indents = (col + 1) :: s.indents }
+    State { s | indents = (col - 1) :: s.indents }
```
Janiczek, line 38: Motivating example:
```elm
case f of
  True -> 1
  False -> 2
```
The case..of parser was at the letter `T`, with location `{row = 2, column = 3 }`.
You can see there is indentation of 2 spaces in the snippet.
Before the fix, `pushIndent` was reading that as indentation of 4 spaces instead of 2.

jfmengels:
Thank you so much @Janiczek :pray:

PR #213, opened 2024-06-22
Add a simple elm-review rule for formatting ranges

jiegillet:
As I was playing around in the codebase, I was often copy/pasting the output of tests into actual tests, and I realized that the ranges/locations are shown in alphabetical order, which is the least readable order possible.

Instead of changing it by hand, I thought I would write a simple `elm-review` rule, so here it is.

### DIFF
--- a/review/src/OrderedRanges.elm
+++ b/review/src/OrderedRanges.elm
@@ -1,0 +1,57 @@
+module OrderedRanges exposing (rule)
+
+import Elm.Syntax.Expression as Expression exposing (Expression, RecordSetter)
+import Elm.Syntax.Node as Node exposing (Node(..))
+import Elm.Syntax.Range exposing (Range)
+import Elm.Writer
+import Review.Fix exposing (Fix)
+import Review.Rule as Rule exposing (Error, Rule)
+
+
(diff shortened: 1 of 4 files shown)


jfmengels:
Aha! This is great! :tada: 

I used to do this with a find+replace regex in my IDE, and I had to this in a LOT of places (because until a while ago all ranges in tests were replaced by `Range.emptyRange`). I wrote the regex in this commit's description 427683dc229c80c5ee111f31c7a1e5f0d1a09085

But this is much better, thanks!

I'll likely rebase the `pratt-parser` branch soon. Let me know if that creates any trouble for you, I'll be able to jump in if/where needed to help you out.

jfmengels:
I'll likely wait until I merge #212, to avoid any problems. Feel free to **not** re-arrange your ranges in that branch until then.

jiegillet:
When you have an `elm-review`, everything is a rule ;)

> I'll likely wait until we merge https://github.com/stil4m/elm-syntax/pull/212, to avoid any problems. Feel free to not re-arrange your ranges in that branch until then.

Sounds good!

PR #214, opened 2024-06-30
Finish Pratt Parser implementation

jiegillet:
This PR makes all the tests pass.

I ended up reverting to the initial layout validations, I needed to tweak the case and let block parsers and it ended up working for the most part.

I did need to change some of the tests that had 1.18-era operators like `!`. This is technically a breaking change, so maybe Pratt parsers is better suited for V8?
Putting them back in is also an option but...

I don't consider this PR final, I think should add more tests at the very least.


src/Pratt.elm
```diff
@@ -134,27 +134,12 @@ expression :
 expression config =
     subExpression 0 <|
         Config
-            { oneOf = List.map failOnIncorrectIndentation config.oneOf
+            { oneOf = config.oneOf
```
jfmengels, line 139: This whole record is now plainly `config`, right?

src/Elm/Parser/State.elm
```diff
@@ -18,9 +18,9 @@ emptyState =
         }
 
 
-storedColumns : State -> List Int
+storedColumns : State -> Maybe Int
 storedColumns (State { indents }) =
-    indents
+    List.head indents
```
jfmengels, line 23: Let's rename this function, because its name doesn't match its return value anymore.

src/Combine.elm
```diff
@@ -205,11 +205,12 @@ manyWithEndLocationForLastElement defaultRange getRange (Parser p) =
...
                 [ p oldState
                     |> Core.map (\( newState, item ) -> Core.Loop ( newState, item :: items ))
-                , Core.succeed ()
-                    |> Core.map
-                        (\() ->
-                            Core.Done ( oldState, ( endLocationForList defaultRange getRange items, List.reverse items ) )
+                , Core.succeed
+                    (Core.Done
+                        ( oldState
+                        , ( endLocationForList defaultRange getRange items, List.reverse items )
```
jfmengels, line 210: I think the `map` on `Core.succeed ()` was done to avoid doing `List.reverse items` eagerly/unnecessarily.

jfmengels (commented):
This is amazing work @jiegillet, thank you so much!

> I did need to change some of the tests that had 1.18-era operators like !. This is technically a breaking change, so maybe Pratt parsers is better suited for V8?

This major version of `elm-syntax` only supports Elm 0.19 which doesn't support custom operators and dropped some. It is not technically a "syntax" error, but it ultimately always leads to a compiler error.
I would rather say that not supporting 0.18 and custom operators is closer to a fix for a false negative rather than a breaking change. Happy to discuss this further, but I feel like I'd need some practical use-cases that would become unsupported to be convinced.

src/Elm/Parser/Tokens.elm
```diff
@@ -261,21 +260,11 @@ excludedOperators =
 
 allowedOperatorTokens : List Char
 allowedOperatorTokens =
-    [ '+', '-', ':', '/', '*', '>', '<', '=', '/', '&', '^', '%', '|', '!', '.', '#', '$', '≡', '~', '?', '@' ]
-
-
-allowedPrefixOperatorTokens : List Char
-allowedPrefixOperatorTokens =
-    ',' :: allowedOperatorTokens
+    [ '+', '-', ':', '/', '*', '>', '<', '=', '&', '^', '%', '|', '!', '.', '?' ]
```
jiegillet, line 263: I left the `%` in there because the compile does it
https://github.com/elm/compiler/blob/2f6dd29258e880dbb7effd57a829a0470d8da48b/compiler/src/Parse/Symbol.hs#L77-L79
but I'm not sure if we should not remove it. Is it used in any operator I forgot about?

jiegillet: Ah, it could be to give better error messages:
```elm
> 1 % 2
-- UNKNOWN OPERATOR ------------------------------------------------------- REPL

Elm does not use (%) as the remainder operator:

3|   1 % 2
       ^
If you want the behavior of (%) like in JavaScript, switch to:
<https://package.elm-lang.org/packages/elm/core/latest/Basics#remainderBy>

If you want modular arithmetic like in math, switch to:
<https://package.elm-lang.org/packages/elm/core/latest/Basics#modBy>

The difference is how things work when negative numbers are involved.
```

jfmengels: Hmm, I'd say we remove it. Maybe we can re-introduce it later if we decide to give better error messages ourselves.

It looks like `(!)`, `(.)` still get marked as valid operators. Would removing their support go through updating this list?
(I won't mind having this part refined in a later PR, doesn't have to be done here).

jiegillet: It's a bit more complicated than that. `(.)` and `(!)` are valid, but I think so are combinations like `(.!.)`, I think it's a remnant of the 1.18 where you could define operators from a list of valid characters.
In this codebase, it's used for parsing operators in expressions but it's also used for parsing the `infix` declaration that the core packages can still use.

But yes, I think we should fix those by replacing `operatorTokenFromList` by a list of known operators, maybe in a follow up PR, this one is getting crowded.

jfmengels:
Is there anything you'd like to add to this PR? I am happy to merge this as is otherwise.

jiegillet:
About a million things 😄 

I think this one is fine for now, there will be more.

jfmengels:
That's absolutely fine. Thank you so much for the help and keep them coming as long as you have fun with this :blush:

PR #216, opened 2024-07-06
Fixes for field access vs application precedence

jiegillet:
Thank you for finding them @jfmengels

How did you find those regressions? I suspect there are more, or at least I'd like to be convinced there are not :)

Maybe we could write a quick and dirty script to compare the ASTs with different versions and run it on some codebases.

### DIFF
--- a/src/Elm/Parser/Expression.elm
+++ b/src/Elm/Parser/Expression.elm
@@ -179,7 +179,7 @@ recordAccessParser =
 
 functionCall : Pratt.Config s (Node Expression) -> ( Int, Node Expression -> Parser s (Node Expression) )
 functionCall =
-    Pratt.infixLeft 95
+    Pratt.infixLeft 90
         (Combine.succeed ())
         (\((Node leftRange leftValue) as left) right ->
             case leftValue of
@@ -466,7 +466,7 @@ ifBlockExpression config =
 
 negationOperation : Config s (Node Expression) -> Parser s (Node Expression)
 negationOperation =
-    Pratt.prefix 9
+    Pratt.prefix 95
         minusNotFollowedBySpaceOrComment
         (\((Node { start, end } _) as subExpr) ->
             Node
--- a/src/Elm/Parser/Expression.elm
+++ b/src/Elm/Parser/Expression.elm
@@ -467,7 +467,7 @@ ifBlockExpression config =
 negationOperation : Config s (Node Expression) -> Parser s (Node Expression)
 negationOperation =
     Pratt.prefix 9
-        minusNotFollowedBySpace
+        minusNotFollowedBySpaceOrComment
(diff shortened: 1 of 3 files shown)


jiegillet:
Well, I wrote the script I was talking about:


```sh
#!/bin/bash

FILES=$@
TARGET="tests/RegressionTests.elm"
PRATT_BRANCH="pratt-parser-regressions"

mkdir -p regressions

for FILE in $FILES
do
    cat > $TARGET << EOF
module RegressionTests exposing (..)

import Elm.Parser
import Expect
import Test


test =
    Test.only <|
        Test.test "parse all" <|
            \() ->
                """
$(cat $FILE | sed 's/\\/\\\\/g' | sed 's/\"\"\"/\\"\\"\\"/g')
"""
    |> Elm.Parser.parseToFile
    |> Debug.log "__RESULT__"
    |> Expect.ok
EOF

    echo "Checking $FILE"

    git checkout $PRATT_BRANCH > /dev/null 2>&1
    npx elm-test | grep __RESULT__ > pratt_result.tmp

    git checkout master > /dev/null 2>&1
    npx elm-test | grep __RESULT__ > master_result.tmp

    diff master_result.tmp pratt_result.tmp > /dev/null
    if [ $? -ne 0 ] ;then
        echo "❌❌❌ Found a parsing difference in $FILE"
        cp $FILE regressions
    fi
done
```

Used with `./find_regressions.sh src/**/*.elm`

It REALLY is quick and dirty, there are false positives when the files are too long (maybe because `Debug.log` has a cutoff, although the length can vary from run to run), but it copies problematic files in `regressions` for further analysis.

I found some regressions, I'll open another PR.

jfmengels:
Thank you for the change!

The way I tested is by replacing my local version of `stil4m/elm-syntax` in my `ELM_HOME` by a symbolic link to the repository. I did that in `elm-review` and ran `elm-test` there.

If your project uses version `7.3.2` of `elm-syntax`, then you can do
```bash
# Need to remove the version first
rm -r ~/.elm/0.19.1/packages/stil4m/elm-syntax/7.3.2
ln -s <your-path-to->/elm-syntax ~/.elm/0.19.1/packages/stil4m/elm-syntax/7.3.2
```
If you change the `elm-syntax` code between two runs of `elm-test` (or something else), I think it's necessary to delete the `elm-stuff` folder (or at least some parts of it, if you want to go faster) otherwise it will reuse a previously cached version of the library.

With this, I also think it's possible to run `elm-review` with the new parser. Though debugging newly reported errors will not be 
a great experience. I think you're right in trying to diff between two versions of parsed output.

I think you could also do your script using [`git-worktree`](https://git-scm.com/docs/git-worktree), which allows you to have multiple branches on the same machine (but I'm not sufficiently used to it for now to give you more advice than this).

PR #217, opened 2024-07-06
Add test failing on master

jiegillet:
After I ran my dirty script on all elm files in `~/.elm/0.19.1/packages/elm` and `~/.elm/0.19.1/packages/jfmengels` and some other files, I only found one regression, and it's pretty funny that it turns out to be a bug on master :)

I'm sure that it was not found before because `elm-format` changes `a-b` to `a - b`, and hardly anyone doesn't use `elm-format`.

### DIFF
--- a/tests/Elm/Parser/ExpressionTests.elm
+++ b/tests/Elm/Parser/ExpressionTests.elm
@@ -388,6 +388,17 @@ all =
                                 ]
                             )
                         )
+        , test "subtraction without spaces" <|
+            \() ->
+                "2-1"
+                    |> expectAst
+                        (Node { end = { column = 4, row = 1 }, start = { column = 1, row = 1 } }
+                            (OperatorApplication "-"
+                                Left
+                                (Node { end = { column = 2, row = 1 }, start = { column = 1, row = 1 } } (Integer 2))
+                                (Node { end = { column = 4, row = 1 }, start = { column = 3, row = 1 } } (Integer 1))
+                            )
+                        )
         , test "negated expression for value" <|
             \() ->
                 "-x"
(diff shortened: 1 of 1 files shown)


jfmengels:
Aha, good catch! :smile: 

Let me know if/when you think we're feeling close to parity.
So far, apart from regressions I found (and were fixed in #216) I've mostly noticed cases where the new version is stricter (but correctly so) in the indentation for let declarations.

jfmengels:
I guess it also means that we can now compare the performance of both approaches. I really hope it's faster :crossed_fingers: 
(But not entirely sure how to test it)

jiegillet:
After today, I feel much better about parity, I'm sure there are edge cases laying here or there, but that's unavoidable :)

Yes, it would be nice to compare performance somehow, my gut feeling is that it's not going to be much faster, but let's test it if we can.

jiegillet:
I gave `elm-explorations/benchmark` a go.

Unfortunately, it's unfathomably slow, it takes more than one hour for parsing a medium size piece of code. I created this file
```elm
-- src/Benchmark/Main.elm
module Benchmark.Main exposing (..)

import Benchmark exposing (Benchmark)
import Benchmark.Runner exposing (BenchmarkProgram)
import Elm.Parser


main : BenchmarkProgram
main =
    Benchmark.Runner.program suite


suite : Benchmark
suite =
    Benchmark.benchmark "Parsing sample file" <|
        \_ -> Elm.Parser.parseToFile sampeFile


sampeFile : String
sampeFile =
    """
module Elm.Parser.Expression exposing (expression, failIfDifferentFrom, functionSignatureFromVarPointer)

-- the first 400 lines the module as a string
```

Then you can run `elm reactor` and run it by navigating on it.

I was patient and I got those two results:
Screenshot 2024-07-07 at 9 36 29
Screenshot 2024-07-06 at 23 34 54

So it looks like the Pratt parser is 15% slower just looking at those numbers.

This kind of makes sense to me, we traded a (presumably fast) post processing of operators for more complex parsing.
The first thing I would try to fix it would be to remove/limit the user of backtrackable code.

PR #219, opened 2024-07-08
Collect documentation comments for declarations while parsing

### LINKED ISSUE #218: Collect documentation comments for declarations while parsing
Currently, all comments are collected in the state. This includes `-- xyz` and `{- xyz -}` comments, but also `{-| xyz -}` comments that are used for declaration comments (top-level declarations only, it's a syntax error anywhere else, probably worth turning into a parsing error).

Then, in the post-processing done in `Elm.Processing`, we get the list of comments from the file, then we loop through the declarations and attach the comment to the declaration. This is wasteful, and we should be able to get rid of this post-processing step.

For context, removing this post-processing step was the main purpose for the `pratt-parser` branch, because re-balancing the tree based around the use of operators was also done here. Now that that is gone, I think we should do the next step. And at that point, I hope we can see a performance improvement over the non-pratt-parser version.

Note: port declarations can't have a documentation comment in the current v7 API, so they should be ignored. We'll want to do that later in the v8 branch where that field has been added.

cc @jiegillet I already had a go at it but was unsuccessful so far. Would you like to have a look at this?

jiegillet:
Closes  #218

This PR collects the documentation at parse time, emptying out the processing module.

Each function declaration now checks for documentation before a function/type declaration (ports will be easy to add).
The one subtlety is not to confuse a module comment

```elm
module Main exposing (..)

{-| so far it looks like module documentation -}
```  

```elm
module Main exposing (..)

{-| so far it looks like a module documentation -}
nope = "it was for a function"
```  
which can happen when there are no imports in between the module docs and first declaration.

### DIFF
--- a/src/Elm/Parser/Comments.elm
+++ b/src/Elm/Parser/Comments.elm
@@ -1,11 +1,12 @@
-module Elm.Parser.Comments exposing (multilineComment, singleLineComment)
+module Elm.Parser.Comments exposing (declarationDocumentation, moduleDocumentation, multilineComment, singleLineComment)
 
 import Combine exposing (Parser, modifyState, string, succeed)
 import Elm.Parser.Node as Node
 import Elm.Parser.State exposing (State, addComment)
 import Elm.Parser.Whitespace exposing (untilNewlineToken)
+import Elm.Syntax.Documentation exposing (Documentation)
 import Elm.Syntax.Node exposing (Node)
-import Parser as Core exposing (Nestable(..))
+import Parser as Core exposing ((|.), Nestable(..))
 
 
 addCommentToState : Parser State (Node String) -> Parser State ()
@@ -31,8 +32,30 @@ multilineCommentInner : Parser State String
 multilineCommentInner =
     Core.getChompedString (Core.multiComment "{-" "-}" Nestable)
         |> Combine.fromCore
+        |> Combine.backtrackable
+        |> Combine.andThen
+            (\comment ->
+                if String.startsWith "{-|" comment then
+                    Combine.fail "unexpected documentation comment"
+
+                else
+                    Combine.fromCore (Core.commit ())
+                        |> Combine.continueWith (Combine.succeed comment)
+            )
 
 
 multilineComment : Parser State ()
 multilineComment =
     parseComment multilineCommentInner
+
+
+moduleDocumentation : Parser State ()
+moduleDocumentation =
+    declarationDocumentation |> addCommentToState
+
+
+declarationDocumentation : Parser State (Node Documentation)
+declarationDocumentation =
+    Core.getChompedString (Core.multiComment "{-|" "-}" Nestable)
+        |> Combine.fromCore
+        |> Node.parser
--- a/src/Elm/Parser/Comments.elm
+++ b/src/Elm/Parser/Comments.elm
@@ -6,7 +6,7 @@ import Elm.Parser.State exposing (State, addComment)
 import Elm.Parser.Whitespace exposing (untilNewlineToken)
 import Elm.Syntax.Documentation exposing (Documentation)
 import Elm.Syntax.Node exposing (Node)
-import Parser as Core exposing ((|.), Nestable(..))
+import Parser as Core exposing (Nestable(..))
 
 
 addCommentToState : Parser State (Node String) -> Parser State ()
--- a/src/Elm/Parser/Declarations.elm
+++ b/src/Elm/Parser/Declarations.elm
@@ -6,7 +6,7 @@ import Elm.Parser.Expression exposing (expression, failIfDifferentFrom, function
 import Elm.Parser.Layout as Layout
 import Elm.Parser.Node as Node
 import Elm.Parser.Patterns exposing (pattern)
-import Elm.Parser.State as State exposing (State)
+import Elm.Parser.State exposing (State)
 import Elm.Parser.Tokens exposing (functionName, portToken, prefixOperatorToken)
 import Elm.Parser.TypeAnnotation exposing (typeAnnotation)
 import Elm.Parser.Typings exposing (typeDefinition)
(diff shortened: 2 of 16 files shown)


src/Elm/Parser/Comments.elm
```diff
@@ -31,8 +32,30 @@ multilineCommentInner : Parser State String
 multilineCommentInner =
     Core.getChompedString (Core.multiComment "{-" "-}" Nestable)
         |> Combine.fromCore
+        |> Combine.backtrackable
+        |> Combine.andThen
+            (\comment ->
+                if String.startsWith "{-|" comment then
+                    Combine.fail "unexpected documentation comment"
```
jiegillet, line 39: I used another `backtrackable` here, I'm not sure how much it affects performance.

src/Elm/Parser/State.elm
```diff
@@ -15,9 +29,20 @@ emptyState =
...
         }
 
 
+parsedImportOrDeclaration : State -> State
+parsedImportOrDeclaration (State state) =
+    State { state | importOrDeclarationParsed = True }
+
+
+checkParsedImportOrDeclaration : State -> Bool
+checkParsedImportOrDeclaration (State { importOrDeclarationParsed }) =
```
jiegillet, line 42: I don't really like the function names I came up with here, suggestions welcome.

src/Elm/Processing.elm
```diff
@@ -67,118 +63,7 @@ addDependency dep (ProcessContext x) =
...
-                        -- Aborting because the next comment can't match the next declaration
-                        ( previousComments, Nothing, comment :: restOfComments )
-
-                LT ->
-                    findDocumentationForRange range restOfComments (comment :: previousComments)
-
-                GT ->
-                    -- Aborting because we went too far
-                    ( previousComments, Nothing, comment :: restOfComments )
+    file
```
jiegillet, line 69: This function is a bit awkward now, it basically does nothing :)

jfmengels:
> The one subtlety is not to confuse a module comment

Actually, the first documentation comment (when there is no import) is the module documentation.

```elm
module A exposing (b)

{-| X

@docs b

-}
type X
    = X


{-| b
-}
b : X
b =
    X
```

When I run `elm-doc-preview` on a package with this module exposed, the first comment is considered to be for the module and not for `X`. I don't know what our current behavior is like, but this is the correct one.

What pointed me in that direction is that `elm-format` re-formats the code above in a way that detaches the comment from the first declaration.

This should simplify the parser quite a bit, no?

src/Elm/Processing.elm
jfmengels: And that's great :grin: 
I'll deprecate it at some point, and it will be removed at the next major version :muscle:

src/Elm/Parser/Declarations.elm
```diff
@@ -21,18 +23,38 @@ declaration : Parser State (Node Declaration)
...
+function : Maybe (Node Documentation) -> Parser State (Node Declaration)
+function maybeDoc =
     Node.parser functionName
         |> Combine.ignore (maybe Layout.layout)
         |> Combine.andThen functionWithNameNode
-        |> Combine.map (\f -> Node (Expression.functionRange f) (Declaration.FunctionDeclaration f))
+        |> Combine.map
+            (\f ->
+                Node
+                    { start = maybeDoc |> Maybe.map (Node.range >> .start) |> Maybe.withDefault (Expression.functionRange f).start
```
jfmengels, line 53: Let's use a case expression instead of `Maybe.map+Maybe.withDefault`, that way we don't do compute `(Expression.functionRange f).start` unnecessarily.

jfmengels: Actually, we already compute it for the `end` field, so let's extract it to a variable so that we only compute it once.

src/Elm/Parser/Declarations.elm
```diff
@@ -21,18 +23,38 @@ declaration : Parser State (Node Declaration)
 declaration =
     oneOf
         [ infixDeclaration
-        , function
-        , typeDefinition
         , portDeclaration
+        , maybeDocumentation
```
jfmengels, line 27: Do you know if we have tests that parse an port declaration with and without a module documentation?

I don't think `infix` declarations can have module documentation (it's done through the function they alias, weirdly)

jiegillet: There are some tests for undocumented  port declarations but we don't have any test for a port with documentation.
It'll make more sense to add one when we support it.

jiegillet:
> This should simplify the parser quite a bit, no?

Yes indeed, it's much cleaner now.

I gave the benchmark another try. I changed the benchmark code so that it just runs a set number of times (without the analysis and the cache warming and all, which makes it less reliable, but not take forever). 
In essence (minus the `Browser.element`)

```elm
sampleSize =
    100

run =
    Task.attempt GotSample (Benchmark.LowLevel.sample sampleSize operation)

operation =
    Benchmark.LowLevel.operation
        (\() -> Elm.Parser.parseToFile sampleFile)
```

Then I ran the benchmark on `jfmengels/elm-review-simplify/2.1.2/src/Simplify.elm` which made an impression on me once by making VSCode repeatedly crash because it couldn't handle the sheer size of it (🤣).
On master I get: `136830 ms to do 100 runs => 1368 ms/run`
On this branch I get `193397 ms to do 100 runs => 1934 ms/run`

So 40% slower 😢
We should try to optimize it for a bit

src/Elm/Parser/Declarations.elm
jfmengels: Right. What concerns me a bit now is that this becomes a parsing error, so I'd like to make sure that isn't the case.

jfmengels:
> I gave the benchmark another try

Ugh. And this is against the latest released version? Yikes...
We do have a few `backtrackable`s but still...
Anyway, that's not a problem for this PR. I think this looks good, I'll add a test for the port later, and try to find some micro optimizations.

jfmengels:
@jiegillet Could you maybe make this benchmark available in a branch somewhere?

PR #221, opened 2024-07-09
[Partt Parser] Basic benchmarking tool

jiegillet:
This is just a hack, not mean to be merged.

Run `elm reactor` and navigate to http://localhost:8000/src/Benchmark/Main.elm
100 runs will take about 3 minutes.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -31,9 +31,12 @@
     ],
     "elm-version": "0.19.0 <= v < 0.20.0",
     "dependencies": {
+        "elm/browser": "1.0.2 <= v < 2.0.0",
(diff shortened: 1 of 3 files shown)


jfmengels:
Btw, a better way to run these benchmarks is to do the following:
```
elm make src/Benchmark/Main.elm --optimize
```
then open the resulting `index.html` file. When benchmarking, you should always use `--optimize` (which I don't think you can do with `elm reactor`?)

jiegillet:
Oh yeah, big difference :)
Yes, that could explain the difference.

jfmengels:
I ended up merging/pushing both commits in some way or another. These are really nice to have by the way :+1:

PR #222, opened 2024-07-10
[Pratt Parser] Optimize parser

jiegillet:
WIP

master: 136830 ms to do 100 runs => 1368 ms/run

As of commit:
- 03d18c8 (head of pratt-parser): 186695 ms to do 100 runs => 1866.95 ms/run
- 883bacf (precise list of infix operators): 183531 ms to do 100 runs => 1835.31 ms/run
- 8ceeb9d (remove confusing off by one indent): 173865 ms to do 100 runs => 1738.65 ms/run
- c27fef3 (minus sign cannot be confused with a comment anymore):  166139 ms to do 100 runs => 1661.39 ms/run 
- cbbc2c6 (optimize multiline comment parser): 152409 ms to do 100 runs => 1524.09 ms/run


tests/Elm/Parser/TokenTests.elm
```diff
@@ -79,10 +79,6 @@ all =
             \() ->
                 parse "=" Parser.prefixOperatorToken
                     |> Expect.equal Nothing
-        , test "operatorToken 15" <|
-            \() ->
-                parse "?" Parser.prefixOperatorToken
-                    |> Expect.equal (Just "?")
```
jfmengels, line 85: We could keep these tests but have them expect an error.

src/Combine.elm
```diff
@@ -203,11 +203,7 @@ manyWithEndLocationForLastElement defaultRange getRange (Parser p) =
             Core.oneOf
                 [ p oldState
                     |> Core.map (\( newState, item ) -> Core.Loop ( newState, item :: items ))
-                , Core.succeed ()
-                    |> Core.map
-                        (\() ->
-                            Core.Done ( oldState, ( endLocationForList defaultRange getRange items, List.reverse items ) )
-                        )
+                , Core.succeed (Core.Done ( oldState, ( endLocationForList defaultRange getRange items, List.reverse items ) ))
```
jfmengels, line 206: This should stay as it was, so that we don't do `List.reverse` eagerly.

src/Elm/Parser/Base.elm
```diff
@@ -16,12 +16,10 @@ typeIndicator =
...
             Combine.oneOf
-                [ Combine.succeed identity
-                    |> Combine.ignore (string ".")
-                    |> Combine.keep Tokens.typeName
+                [ string "."
+                    |> Combine.continueWith Tokens.typeName
                     |> Combine.andThen (\t -> helper (typeOrSegment :: moduleNameSoFar) t)
-                , Combine.succeed ()
-                    |> Combine.map (\() -> ( List.reverse moduleNameSoFar, typeOrSegment ))
+                , Combine.succeed ( List.reverse moduleNameSoFar, typeOrSegment )
```
jfmengels, line 22: This should stay as it was, so that we don't do List.reverse eagerly.

src/Elm/Parser/Expression.elm
```diff
@@ -25,60 +25,56 @@ expression =
...
+            -- TODO Add tests for all operators
             -- TODO Report a syntax error when encountering multiple of the comparison operators
             -- `a < b < c` is not valid Elm syntax
-            -- TODO Add tests for all operators
-            , infixNonAssociative 4 "=="
-            , infixNonAssociative 4 "/="
-            , infixNonAssociative 4 "<="
-            , infixNonAssociative 4 ">="
+            [ recordAccess
+            , infixLeft 1 "|>"
```
jfmengels, line 54: A lot of these operators start with a shared symbol, notably `|` and `<`. I'm wondering whether we have a parser a bit like
```elm
Parser.succeed identity
  |. Parser.symbol "|"
  |= Parser.oneOf
    [ Parser.succeed ( Left, 1 )
        |. Parser.symbol ">"
    , Parser.succeed ( Right, 2 )
        |. Parser.symbol "|"
    , ...
    ]
```

From what I can tell this is a bit involved because the `andThenOneOf` is supposed to have static precedence in order to filter them out based on the current precedence (which might actually be quite slow as well?)

(I'm just brainstorming here)

jiegillet: Worth exploring, if the precedence is the same, then it could work. If it's different, I'm not sure...

src/Elm/Parser/Expression.elm
```diff
@@ -24,67 +24,71 @@ expression : Parser State (Node Expression)
...
+            , lambdaExpression
+            , letExpression
+            , ifBlockExpression
+            , recordAccessFunctionExpression
+                |> Pratt.literal
+            , negationOperation
+            , charLiteralExpression
+                |> Pratt.literal
             ]
+                |> List.map (\parser config -> Layout.positivelyIndented |> Combine.continueWith (parser config))
```
jiegillet, line 48: This change (3038846) is actually terrible in terms of optimization, it makes the parser quite a bit slower, so I think I will revert it for now.
It makes the parsing more robust, by preventing things like `foo = a\n+b`, which might be worth doing, but it's a but counterproductive in the context of this PR.

src/Elm/Parser/Base.elm
```diff
@@ -16,10 +16,12 @@ typeIndicator =
         helper : ModuleName -> String -> Parser s ( ModuleName, String )
         helper moduleNameSoFar typeOrSegment =
             Combine.oneOf
-                [ string "."
-                    |> Combine.continueWith Tokens.typeName
+                [ Combine.succeed identity
+                    |> Combine.ignore (string ".")
+                    |> Combine.keep Tokens.typeName
```
jfmengels: I think we can re-apply this reverted change :+1:

src/Elm/Parser/Expression.elm
```diff
@@ -282,7 +282,7 @@ recordContents config =
 
 recordUpdateSyntaxParser : Config State (Node Expression) -> Node String -> Parser State Expression
 recordUpdateSyntaxParser config fname =
-    Combine.succeed (\e -> RecordUpdateExpression fname e)
+    Combine.succeed (RecordUpdateExpression fname)
```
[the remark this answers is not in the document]
jiegillet: Oh really! I went with my gut there, but I should have done my research.
I think I'll stop working on this right PR for a little bit, my last few commits are shaky and if you are working on it at the same time, it's counterproductive. Please go ahead and commit your changes :)

src/Elm/Parser/Expression.elm
jfmengels: Alright, sounds good :+1:

src/Elm/Parser/Expression.elm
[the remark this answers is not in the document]
jfmengels: No worries. Shall I merge what you have already? (it's a draft but I can get merge it anyways if you want).
You've done some really nice improvements, thank you!

> I should have done my research

No worries. This is one of the things that if you don't know you'll have a really hard time figuring out :sweat_smile: 
(and it usually doesn't matter, unless you really want to crank out performance)

jfmengels:
Thank you so much, this is a really nice performance increase!

PR #223, opened 2024-07-16
Performance improvements to the Parser

jfmengels:
Follow up to #222

(Metrics are for my machine, not directly comparable to @jiegillet's results)

Before pratt-parser: 73400.5 ms to do 100 runs => 734.005 ms/run (weird, @jiegillet's results for this was much slower?)
Before: 146231.60000002384 ms to do 100 runs => 1462.3160000002383 ms/run
After: 101167.20000004768 ms to do 100 runs => 1011.6720000004768 ms/run

There are plenty of optimizations done here:
- Improve the arity of functions (see https://blogg.bekk.no/how-elm-functions-work-71cab7426a2f), which leads to some odd code in applicatives
- Trying to avoid using `Combine` as much as possible, and using the core parser as much as possible instead, wrapping as late as possible.
	- `Combine` has a lot of overhead, so sticking to using core parser is much better.
- Try to avoid unnecessary collection of data for things that get ignored (this is now enforced when using `Combine.ignore` and friends)

### DIFF
--- a/review/suppressed/NoImportingEverything.json
+++ b/review/suppressed/NoImportingEverything.json
@@ -3,7 +3,6 @@
   "automatically created by": "elm-review suppress",
   "learn more": "elm-review suppress --help",
   "suppressions": [
-    { "count": 13, "filePath": "src/Elm/Writer.elm" },
-    { "count": 1, "filePath": "src/Elm/Parser/TypeAnnotation.elm" }
+    { "count": 13, "filePath": "src/Elm/Writer.elm" }
   ]
 }
--- a/src/Combine.elm
+++ b/src/Combine.elm
@@ -382,10 +382,13 @@ sepBy1WithoutReverse sep p =
 
 
 between : String -> String -> Parser state a -> Parser state a
-between lp rp p =
-    symbol lp
-        |> andThen (\() -> p)
-        |> ignore (symbol rp)
+between lp rp (Parser p) =
+    Parser <|
+        \state ->
+            Core.succeed identity
+                |. Core.symbol lp
+                |= p state
+                |. Core.symbol rp
 
 
 parens : Parser state a -> Parser state a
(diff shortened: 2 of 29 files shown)


jiegillet (commented):
Very very cool! So many things I would not have thought of :)

jfmengels:
Alright, I think I'm done with this batch. I know I've introduced some complexity (I'm mostly thinking all the core->combine and combine->core functions). If anyone works with this, feel free to change or refactor it. This was an incremental change

Here are my most recent benchmarks:
- Master: 970 ms/run
- Before: 1543 ms/run
- After: 1065 ms/run

In other words, we're pretty close to the speed from before, although I'm surprised we're not faster. @jiegillet could you try running these benchmarks on your device?

@jiegillet If you feel like reviewing, please do, but it's also quite alright to say you don't want to look at...  *checks notes* ... 125 commits :sweat_smile:

jiegillet:
Very cool!
I didn't check commit-by-commit, but it overall looks good. 
Some of the changes did make the code more complex, but it's hard to avoid for an optimization PR.

as for benchmarks on my machine:
master: 1374 ms/run
before: 1579 ms/run
after: 1262 ms/run

very impressive! Although these numbers are pretty unreliable, this is definitely on par with master!

jiegillet (approved):

jfmengels:
@jiegillet do you mean to say that for you this branch (not including the other PRs up until #225) is already faster than master? That's great if it is, though not sure how we get such different benchmarks.

jiegillet:
Yes, that's what I got.
I guess 100 runs is really not reliable, one bad run makes the average much longer. The time to run with the default settings is like an hour and a half after all :)
I wouldn't worry about it, we both see that it's faster, and the code is still readable, it's great that you found even more ways to optimize the Pratt parser.

PR #224, opened 2024-07-16
Compute the precedence before passing the config to the parser

jfmengels:
This makes it so that we can check the precedence of something, to check whether we need to evaluate **before** we create the prser by calling it with the config.

This reaches 822 ms/run! That's about 20% faster. And this should now be faster than than `master`!

To me, this very much indicates that creating the parsers over and over again is a big slowdown.
One annoying thing with the current state of the pratt parser, is that we pass the `Config` to the parsers, which means we re-evaluate it every time.

```elm
listExpression : Config State (Node Expression) -> Parser State (Node Expression)
listExpression config =
    Combine.succeed ListExpr
        |> Combine.ignoreEntirely squareStart
        |> Combine.ignore (Combine.maybeIgnore Layout.layout)
        |> Combine.keep (Combine.sepBy "," (Pratt.subExpression 0 config))
        |> Combine.ignoreEntirely squareEnd
        |> Node.parser
```

This gets re-created over and over inside of `Pratt.subExpression`, which gets called loads of times.

Maybe we could memoize the config somehow, or move it to inside the `State`. I don't know yet.

### DIFF
--- a/src/Pratt.elm
+++ b/src/Pratt.elm
@@ -200,7 +200,7 @@ subExpression currentPrecedence ((Config conf) as config) =
                 )
             )
         |> Combine.andThen
-            (\leftExpression -> Combine.loop leftExpression (expressionHelp config currentPrecedence))
+            (\leftExpression -> Combine.loop leftExpression (\expr -> expressionHelp config currentPrecedence expr))
 
 
 {-| This is the core of the Pratt parser algorithm.
--- a/src/Pratt.elm
+++ b/src/Pratt.elm
@@ -191,7 +191,7 @@ A parser for sub-expressions between parentheses like this:
 
 -}
 subExpression : Int -> Config state expr -> Parser state expr
-subExpression precedence ((Config conf) as config) =
+subExpression currentPrecedence ((Config conf) as config) =
     conf.spaces
         |> Combine.continueWith
             (Combine.lazy
@@ -200,7 +200,7 @@ subExpression precedence ((Config conf) as config) =
                 )
             )
         |> Combine.andThen
-            (\leftExpression -> Combine.loop leftExpression (expressionHelp config precedence))
+            (\leftExpression -> Combine.loop leftExpression (expressionHelp config currentPrecedence))
 
 
 {-| This is the core of the Pratt parser algorithm.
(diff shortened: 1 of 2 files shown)


src/Pratt.elm
```diff
@@ -215,31 +215,29 @@ Also note that `oneOf` and `andThenOneOf` parsers may call recursively
...
     conf.andThenOneOf
-        |> List.filterMap (\toOperation -> filter (toOperation config) precedence leftExpression)
-        |> Combine.oneOf
-
-
-filter : ( Int, e -> Parser state e ) -> Int -> e -> Maybe (Parser state e)
-filter ( precedence, parser ) currentPrecedence leftExpression =
-    if precedence > currentPrecedence then
-        Just (parser leftExpression)
+        |> List.filterMap
```
lue-bird, line 232: I don't know how often this is called, but (as you know) most `List.` helpers aren't well optimized.
For e.g. `filterMap`, a fold or recursion version is twice as fast at least (more if you don't care about resulting order).
image
```elm
listJustsToAnyOrderUsingRecursion : (element -> Maybe value) -> List element -> List value
listJustsToAnyOrderUsingRecursion elementToMaybe list =
    listJustsToReverseOrderUsingRecursionFrom [] elementToMaybe list


listJustsToReverseOrderUsingRecursionFrom : List value -> (element -> Maybe value) -> List element -> List value
listJustsToReverseOrderUsingRecursionFrom soFar elementToMaybe list =
    case list of
        [] ->
            soFar

        head :: tail ->
            listJustsToReverseOrderUsingRecursionFrom
                (case elementToMaybe head of
                    Nothing ->
                        soFar

                    Just value ->
                        value :: soFar
                )
                elementToMaybe
                tail
```

jfmengels: You're right. In elm-review however, I patch these core functions to be the same as the one in elm-optimize-level-2, where they are fast.
If I use a custom solution, then I can't really patch unfortunately (and it will be slow-ish for all)

I don't know how much I should consider this not running through elm-review 😅

jiegillet: > For e.g. filterMap, a fold or recursion version can twice as fast (more if you don't care about resulting order).

in this case, we care about the order because the most used parsers are on top. Although we could simply reverse the original list :)

jiegillet (commented):
I'm not seeing the 20% improvement you mentioned, if anything it seems a bit slower (although not significantly).

This time I ran the code using `--optimize`, and I also used a different browser that was only doing that (not sure if it matters or not), and I also ran it several times.

master: 563 - 569 - 581 ms/run
before: 647 - 648 - 647 ms/run
after: 650 - 652 - 684 ms/run

after #225: 537 - 552 - 565 ms/run

jfmengels:
Alright. Thank you very much for looking into this @jiegillet @lue-bird.

I think that maybe we're good for releasing this...? Do you know if/how we should do additional checks?

jiegillet:
I ran the regression finder on a bunch of packages, I haven't found anything else than the `n-1` bug on master.
So... I think we're good, exciting 😸

PR #226, opened 2024-07-18
[Pratt parser] handle nested comments

jiegillet:
I ran the regression script on many Elm packages and found one.
I'm pretty confident it's the last one.

I don't particularly like the way we are parsing comments right now, but looking ahead without consuming is required here, and I'm not sure how to handle that best, so I'm open to suggestions.


jfmengels (approved):
Thank you for the fix :+1:

PR #227, opened 2024-07-18
Support multi-line string pattern in case expressions

### LINKED ISSUE #134: Multiline string literal in pattern fails to parse
Disclaimer: I hope to never see it in real Elm code, but... :rofl: 

Multiline string literals in patterns are allowed in Elm code but cannot be parsed by `elm-syntax`.  For example, the following module will compile but fails to parse with `elm-syntax`:

```elm
module Foo exposing (..)

tests =
    case "string" of
        """multiline
       pattern
       string""" ->
            "long"

        _ ->
            "not long"
```

This is true even if the literal does not span multiple lines, e.g.

```elm
module Foo exposing (..)

tests =
    case "string" of
        """multiline string""" ->
            "long"

        _ ->
            "not long"
```

jfmengels:
Fixes #134

### DIFF
--- a/src/Elm/Parser/Patterns.elm
+++ b/src/Elm/Parser/Patterns.elm
@@ -141,7 +141,10 @@ unitPattern =
 
 stringPattern : Parser state (Node Pattern)
 stringPattern =
-    Tokens.stringLiteral
+    Core.oneOf
+        [ Tokens.multiLineStringLiteral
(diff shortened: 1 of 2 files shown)


jiegillet (approved):
LGTM. 
I can finally use elm-review on all my codebases with multiline string patterns 🎉

PR #228, opened 2024-07-19
Enable NoUnsortedRecords rule

jfmengels:
This removes the need for the original intent behind the `OrderedRanges` rule (which I've renamed to `AstFormatting`).

I think this is interesting to try out, but happy to disable it if anyone thinks it's painful somehow.


jiegillet (approved):
Sounds good to me. The more automations, the less on my mind :)

PR #229, opened 2024-07-19
More small parse performance improvements

lue-bird:
running the benchmarks with eol2 --optimize-speed in chromium

- before:  81955.5 ms to do 100 runs => 819.555 ms/run
- after: 42913.90000000596 ms to do 100 runs => 429.1390000000596 ms/run

(the improvements are pretty consistent now)

Done by

- moving let declarations to module level if it requires no changes
- changing `succeed () |> map (\() -> x)` to `lazy (\() -> succeed x)`
- changing `succeed a |. x` to `map (\() -> a) x`
- small optimizations for e.g. how string literal parts are folded
- ignore reserved words for type names
- change from `Dict` for `operations` to `case` (and actually pre-compute available operations for all precedences)
- shortcutting lower/uppercase character checks
- faster layout primitives

Not all changes are necessarily performance related. Some change ignored unit arguments to `\() ->`. One fixes import alias parsing.


src/Combine.elm
```diff
@@ -248,11 +248,11 @@ manyWithEndLocationForLastElement defaultRange getRange (Parser p) =
...
                 [ p oldState
                     |> Core.map (\( newState, item ) -> Core.Loop ( newState, item :: items ))
-                , Core.succeed ()
-                    |> Core.map
-                        (\() ->
-                            Core.Done ( oldState, ( endLocationForList defaultRange getRange items, List.reverse items ) )
-                        )
+                , Core.lazy
+                    (\() ->
+                        Core.succeed
```
jfmengels, line 253: Hmm, I wonder if this is faster. As far as I know `Parser.lazy` is only there to avoid cyclic references between parsers, not to improve performance. Maybe it does as well but I'm not sure :thinking:

src/Elm/Parser/Tokens.elm
```diff
@@ -143,28 +143,22 @@ escapedCharValue =
...
-                |. Core.symbol "\\"
-                |= escapedCharValue
-            , Core.getChompedString (Core.chompIf (always True))
-            ]
-        |. Core.symbol "'"
+slashEscapedCharValue : Core.Parser Char
+slashEscapedCharValue =
+    Core.succeed identity
+        |. Core.symbol "\\"
+        |= escapedCharValue
```
jfmengels, line 150: This is much nicer to read :smile:

src/Elm/Parser/Expression.elm
```diff
@@ -133,21 +133,22 @@ functionCall =
...
-
-        end : String
-        end =
-            "|]"
-    in
     Core.mapChompedString
-        (\s () -> s |> String.dropLeft (String.length start) |> GLSLExpression)
-        (Core.multiComment start end NotNestable)
-        |. Core.symbol end
+        (\s () -> s |> String.dropLeft (String.length glslStart) |> GLSLExpression)
```
jfmengels, line 149: We should probably also make `String.length glslStart` a constant, right? :thinking: 
Or should/could we get the chomped string on the parser between start and end (without the dropLeft) :thinking:

src/Combine.elm
lue-bird: map needs to check whether it's valid or not underneath, so this kind of makes sense to me

jfmengels: Oh right, that's a good point. But to do that it also re-defines the parser everytime it gets evaluated, whereas with `map` the parser is already defined, and (from prior optimizations) that was quite bad for performance.

I really don't know what is faster though. I'm okay with taking this change, and then benchmarking this change to know what is faster, because this would be a cool learning.

src/Elm/Parser/Tokens.elm
```diff
@@ -162,92 +162,69 @@ characterLiteral =
...
+    Core.oneOf
+        [ Core.symbol "\"" |> Core.map (\() -> Done (partsSoFarReverse |> List.Extra.listReverseThenStringConcatFast))
+        , Core.succeed (\v -> Loop (String.fromChar v :: partsSoFarReverse))
+            |. Core.symbol "\\"
+            |= escapedCharValue
+        , Core.succeed
+            (\start ->
+                \value ->
+                    \end ->
+                        if start == end then
```
jfmengels, line 183: Can this condition even be `False`? :thinking:

lue-bird: Good eye. I don't think it can be False as well.

src/Elm/Parser/Tokens.elm
```diff
@@ -166,14 +165,14 @@ stringLiteral : Core.Parser String
...
 
-stringLiteralHelper : List String -> Core.Parser (Step (List String) String)
-stringLiteralHelper partsSoFarReverse =
+stringLiteralHelper : String -> Core.Parser (Step String String)
+stringLiteralHelper stringSoFar =
     Core.oneOf
-        [ Core.symbol "\"" |> Core.map (\() -> Done (partsSoFarReverse |> List.Extra.listReverseThenStringConcatFast))
-        , Core.succeed (\v -> Loop (String.fromChar v :: partsSoFarReverse))
+        [ Core.symbol "\"" |> Core.map (\() -> Done stringSoFar)
+        , Core.succeed (\v -> Loop (stringSoFar ++ String.fromChar v))
```
jfmengels, line 175: In practice, maybe we should do `stringSoFar ++ String.fromChar v ++ ""`, that sometimes improves performance.
(The reasoning: `++` can be used for lists and arrays, and by default Elm doesn't know which one, so it uses some kind of `append` function here. If however there is a very explicit string in the `++` chain, then it knows, and uses JS' `+` operator.)

src/Elm/Parser/Tokens.elm
```diff
@@ -197,17 +196,17 @@ multiLineStringLiteral : Core.Parser String
...
+multiLineStringLiteralStep : String -> Core.Parser (Step String String)
+multiLineStringLiteralStep stringSoFar =
     Core.oneOf
         [ Core.symbol "\"\"\""
-            |> Core.map (\() -> Done (List.Extra.listReverseThenStringConcatFast partsSoFarReverse))
+            |> Core.map (\() -> Done stringSoFar)
         , Core.symbol "\""
-            |> Core.mapChompedString (\v () -> Loop (v :: partsSoFarReverse))
-        , Core.succeed (\v -> Loop (String.fromChar v :: partsSoFarReverse))
+            |> Core.mapChompedString (\v () -> Loop (stringSoFar ++ v))
```
jfmengels, line 197: `Core.map (\() -> Loop (stringSoFar ++ "\"") `  should be even faster, right?

(PS: I'm just mentioning the improvements I think of, you stop and mark this as ready for review whenever you want :D )

jfmengels, line 197: `Core.symbol xyz |> Core.mapChompedString (\v () -> ...)` can also be replaced by `Core.string xyz |> Core.map (\v -> ...)` if I'm not mistaken. (Hey, `Simplify` idea :sweat_smile: )

src/Elm/Parser/Tokens.elm
```diff
@@ -218,7 +218,7 @@ typeName =
     Core.variable
         { start = Unicode.isUpper
         , inner = \c -> Unicode.isAlphaNum c || c == '_'
-        , reserved = reservedList
+        , reserved = Set.empty
```
jfmengels, line 221: Oh good catch!

src/Elm/Parser/Tokens.elm
lue-bird: What's Code.string and why am I not finding it...

jfmengels: Oh no, my bad, it was `Combine.string`, and I removed it in a previous PR. Carry on, nothing to see...

src/Elm/Parser/Tokens.elm
lue-bird: on that note, as you've benchmarked, it might be faster to check for reserved words after `variable` has been parsed using a `case` instead of the current `Set`. Do you think that's a reasonable change?

jfmengels: (Did I benchmark that?) Could be reasonable, but I'd imagine a `Set` to be faster. To the benchmarks :smile:

src/Elm/Parser/Patterns.elm
```diff
@@ -125,16 +125,14 @@ qualifiedPatternArg =
...
+    Core.map (\() -> AllPattern) (Core.symbol "_")
         |> Node.parserCore
         |> Combine.fromCore
 
 
 unitPattern : Parser state (Node Pattern)
 unitPattern =
-    Core.succeed UnitPattern
-        |. Core.symbol "()"
+    Core.map (\() -> UnitPattern) (Core.symbol "()")
```
jfmengels, line 135: Is this faster?

src/Elm/Parser/Tokens.elm
lue-bird: I was talking about https://github.com/jfmengels/elm-benchmarks/blob/master/src/WhatIsFaster/SetMemberOrMultipleComparisons-Results-Chrome.png

src/Elm/Parser/Patterns.elm
lue-bird: This seemed to be the default in the rest of the codebase. I imagine this is faster because it's more direct (instead of construct then andMap via |=)

src/Elm/Parser/Tokens.elm
jfmengels: Oh right! Yeah, let's benchmark this with the strings in the reserved list. There are more strings and they're longer so that may have an impact).

lue-bird: Result (ran it twice with the same result): 
image
It's not as much as I'd have hoped but still nice due to how often it's called.

lue-bird: something substantial:
image
by shortcutting `unicode.isAlphaNum` with `Char.isAlphaNum c`. This should be safe, right?

src/Elm/Parser/Patterns.elm
lue-bird: Here's the result (ran twice, always about 10%):
image

src/Elm/Parser/Tokens.elm
jfmengels: Agreed, good idea

src/Elm/Parser/Patterns.elm
jfmengels: Good to know!

src/Elm/Parser/Base.elm
```diff
@@ -5,15 +5,32 @@ import Elm.Parser.Tokens as Tokens
...
+
+
+moduleNameOrEmpty : Core.Parser ModuleName
+moduleNameOrEmpty =
+    Core.oneOf
+        [ Core.succeed listCons
+            |. Tokens.dot
+            |= Tokens.typeName
+            |= Core.lazy (\() -> moduleNameOrEmpty)
+        , Core.lazy (\() -> Core.succeed [])
```
jfmengels, line 30: Using Core.succeed [] is probably fine, seeing as it will only ever be evaluated once.

src/Elm/Parser/Base.elm
```diff
@@ -5,15 +5,32 @@ import Elm.Parser.Tokens as Tokens
...
+    \tail -> head :: tail
+
+
+moduleNameOrEmpty : Core.Parser ModuleName
+moduleNameOrEmpty =
+    Core.oneOf
+        [ Core.succeed listCons
+            |. Tokens.dot
+            |= Tokens.typeName
+            |= Core.lazy (\() -> moduleNameOrEmpty)
```
jfmengels, line 29: Since nothing gets computed in the lazy here, I think it's unnecessary to have Core.lazy.

lue-bird: it's recursive, that's the catch

src/Elm/Parser/Expression.elm
```diff
@@ -667,16 +672,115 @@ operation currentPrecedence leftExpression =
...
+
+
+operation95 : Node Expression -> Parser State (Node Expression)
+operation95 leftExpression =
+    operation 95 leftExpression
+
+
+operation100 : Node Expression -> Parser State (Node Expression)
+operation100 leftExpression =
+    operation 100 leftExpression
```
jiegillet, line 783: How much faster is this? Because it seems a little excessive :)

lue-bird: indeed, benchmarking rn. There are likely more low-hanging fruits (e.g. the List.map inside oneOf there that can be moved into the filterMap one level higher)

lue-bird: benchmarks show predefining these doesn't make a difference, you're right.

lue-bird:
We could do some more random things. For example for
```elm
Core.oneOf
    [ Tokens.multiLineStringLiteral
    , Tokens.stringLiteral
    ]
```
we could share the first " at the cost of readability and a tiny performance improvement.

Same in
```elm
Combine.oneOf
        [ functionWithSignature pointer
        , functionWithoutSignature pointer
        ]
```
where technically both share the declared name at the start, so if the first one fails it shouldn't need to be parsed again, or fail both directly if the name is invalid.
Would changes like that be acceptable?

Another thing: `keyword` (which looks ahead for whitespace) is used for tokens that always have layout after them (e.g. port in port module). Should we switch to token / symbol there? Actually so far I've only seen keyword being used with correct layout after.

src/Elm/Parser/Expression.elm
jfmengels: I think the way it's done right now, we're not actually precomputing anything, and instead recomputing (filter the precedences) every time. I'm slightly in doubt as to whether precomputation worked before this PR either, though I at some believed I had made it work.

If we got rid of the expression argument in `operator` (probably not possible?), or if we did the precedence filtering and then returned a lambda that took the expression argument, then it would work for sure.

jfmengels:
List.filterMap is optimized in a non-merged PR (https://github.com/mdgriffith/elm-optimize-level-2/pull/75). But it does get optimized in elm-review (which is not using eol2, just doing a few similar patches).

Random optimization idea as well (I'll reply to your other questions tomorrow morning): we do a LOT of string concatenations when parsing string literals. But most of the time we can get the contents through a simple get chompedstring. The only times we can't do that is when there's escaping (which is not the common case) and even then we could probably use getChompedString for the rest of the strings, leading to a lot less concatenations.

lue-bird:
I don't think this will improve anything because literals that don't contain escaped characters are already parsed fast by `chompWhile`.

src/Elm/Parser/Expression.elm
lue-bird: are you looking at the latest state where we have module-level filtered lists?
And I also don't think it worked in the `Dict` version because the filterMap is inside the constructed lambda for each entry. But I haven't tested so who knows

jfmengels: Sorry, what do you mean with "module-level filtered lists"?
And yeah, I don't think it worked with `Dict` either, although at the time I did think so. I'll look into it at some point again.

jfmengels:
Yeah, we already do that, I only remembered the part where we add one character. Too bad :sweat_smile: 
(And I had to braindump that before going to bed, didn't have time to re-check)


> we could share the first " at the cost of readability but giving us a tiny performance improvement.

I think that's a good idea. I'm honestly confused as to how this even works, because my mental model says that the first `"` character should cause `multimultiLineStringLiteral` to commit and therefore cause to not be evaluated.

For the other location (`functionWithSignature`, ...), I'm less sure. I think it would be good to have, but at least the way we often do things is by passing the argument (through `andThen` or `map` to a new function which redefines. At this point, I believe that having parser defined as constants leads to faster code, therefore redefining the parser inside of a function leads to slower code. The `pointer` argument is something that I'd like to remove already, and I wonder if sharing the parser for the type name is not going to make that even harder.

> Re:keyword

I don't yet understand the differences between symbol/token/keyword. I'm fine with changing to the fastest one. And maybe having some guidance written down somewhere to explain which one to use in general (or in which case).

> I have a commit ready for replacing record type alias constructor calls with fields. Should I add this here or are there other plans (like: "keep them because records might get replaced by variants in v8")

In terms of breaking changes, I think it's fine. Do you know if it's faster? It removes an intermediary function that does the same thing, but maybe that function would get optimized by the JIT. Those are the questions I pondered and didn't know the answer yet.

> do you use any tools to look at performance at the bigger picture?

I don't have any useful tool for this unfortunately, just individual smaller benchmarks like the ones you made (but which I haven't used recently, do you have one somewhere that I could use as a starting point?) and @jiegillet's 100 run benchmark.

> Core.variable is faster than symbol |. chompWhile

I'm confused as to the final results for this, what is finally the fastest? And is this a pattern we use often? I only see it in `many1Spaces` (which is ran very often yes). And :people_hugging: for your ruined day.

lue-bird:
> Core.variable is faster than symbol |. chompWhile: I'm confused as to the final results for this, what is finally the fastest?

Core.variable is consistently ~15% faster than with chompWhile. I imagine it could even be faster with `oneOf [ variable, succeed "" ]` replacing regular `chompWhile` calls but that I haven't verified.

> do you have one somewhere that I could use as a starting point? and @jiegillet's 100 run benchmark.

I pushed my benchmark code to this branch: https://github.com/lue-bird/elm-syntax/tree/specific-and-full-benchmarks/benchmark

> I don't yet understand the differences between symbol/token/keyword

The doc explains it well: symbol = token and keyword only parses successfully if the character after the symbol is whitespace.

> I'm honestly confused as to how this even works, because my mental model says that the first " character should cause multimultiLineStringLiteral to commit and therefore cause to not be evaluated.

That's the cool thing about `symbol "\"\"\""`. It only commits if the whole thing parses successfully.

src/Elm/Parser/Declarations.elm
```diff
@@ -129,6 +126,78 @@ functionWithNameNode start ((Node _ startName) as startNameNode) maybeDocumentat
         ]
 
 
+letFunctionWithoutDocumentationWithSignatureWithNameAndMaybeLayoutBacktrackable : Parser State (Node Declaration)
+letFunctionWithoutDocumentationWithSignatureWithNameAndMaybeLayoutBacktrackable =
```
jfmengels: That's a mouthful :sweat_smile: 
The name is fine, I'm happy for anyone to rename it later if they find a better name, but... why the `let` at the beginning? This is not for a `let..in` function, right?

lue-bird: got renamed in a later commit

jfmengels: I just saw time, I was going to edit my comment but you were too quick :sweat_smile:

jfmengels:
This has been amazing @lue-bird, I can't believe you've pretty much made the parser twice as fast :exploding_head: 

I don't see any reason not to merge this as it is. Feel free to continue the performance improvements in new PRs, if you still have the performance itch :smile: 

(Note: I've cleaned up the branch and force-pushed it to merge it nicely.)

PR #230, opened 2024-07-20
[v8] Ongoing branch for v8, rebased on 7.3.3

jiegillet:
I ended up doing an normal rebase, it was long, but not complicated.

I was really grateful for simple PRs, with simple descriptions that I could rely on. For example, if it was a rename, I could just ignore all conflicts and rename what needed renaming.

Some commits disappeared because they were doing things that were already done on master (like separating the expression parser in a different file).

I hope I didn't miss anything, especially new tests or something.

Some random notes I took:
- I don't know how to feel about `Tuple`. Grouping them all together makes sense in the context of parsing, but really, `Unit` and `Parenthesized` feel semantically different from tuples (especially `Parenthesized`), and it would be nice to validate that tuples don't have too many fields, so having `Tuple3` and `Tuple4` (or something) would be nice. I haven't much though about how that would make AST analysis easier or not.
- Maybe function calls should be `FunctionCall (Node Expression) (Node Expression) (List (Node Expression))` (instead of `FunctionCall (Node Expression) (List (Node Expression))`) in the sense that function arguments should be a non-empty list.
- ~~`isOperatorApplication` should get renamed `isOperation`~~ (Edit: I just did it)
- `Elm.Internal.RawFile` is not needed anymore


jfmengels:
Thank you!
I've re-rebased your branch on top of (current) `master`, and pushed it to `breaking-changes-v8` (meaning https://github.com/stil4m/elm-syntax/pull/188). @lue-bird was very active so that was a bit hard :sweat_smile: 

I dropped a few commits that were more of the code organization type (like "move signature parser to another module") which I think are fine to drop (and re-do later if we want to).

PR #232, opened 2024-07-22
Better file sample tests

lue-bird:
- a sample file contained a parsing error (it's a bug in elm-syntax that it was parsed in the first place, will fix in the next PR)
- there aren't samples for numbers you might expect, which means going through the list of them using `List.indexedMap` as was done in the file tests gave the wrong sample indexes
- The errors produced by these sample tests were displayed as `Nothing |expectNotEqual Nothing`. Now it shows the actual info for debugging

I thought I was going insane for about an hour due to the combination of these 3 (4) things, so I'm glad this is fixed now.

### DIFF
--- a/tests/Elm/Parser/FileTests.elm
+++ b/tests/Elm/Parser/FileTests.elm
@@ -1,10 +1,12 @@
 module Elm.Parser.FileTests exposing (all)
 
+import Combine
 import Elm.Internal.RawFile as InternalRawFile
 import Elm.Parser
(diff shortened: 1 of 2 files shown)


jfmengels:
Nice, I think at some point we should turn the samples into more regular tests.

lue-bird:
Honestly I like them for what they are. Some more real world snippets that if you're lucky might reveal a case you didn't think of

PR #233, opened 2024-07-23
Even faster

lue-bird:
WIP
- before: 42913.90000000596 ms to do 100 runs => 429.1390000000596 ms/run
- after: 31419.200000000186 ms to do 100 runs => 314.1920000000019 ms/run

done by
- ~more agressive pre-definitions and converting `shared |> Result.andThen (\partial -> succeed (with partial) ...)` to `map (\partial -> \with -> with partial) shared |> keep ...)`~ use intermediate union types for parsers that parse the rest of some structures
- convert `succeed` then drop/keep `parser` in pipelines to `map (\parsed -> ...) parser` (this was pretty fruiitful actually)
- introducing Combine.maybe(Map) as a more performant `oneOf [..., succeed ]`
- switch from `indents : List Int` in `State` to ~`indent : Int`~ `Core.getIndent` and `Core.withIndent`

I can think of some uglier ones if we want to get desperate.


src/Elm/Parser/TypeAnnotation.elm
```diff
@@ -41,27 +36,34 @@ arrowRightToTypeAnnotation =
...
+        [ parensTypeAnnotation
+        , typedTypeAnnotationWithArguments
+        , genericTypeAnnotation
+        , recordTypeAnnotation
+        ]
+
+
+typeAnnotationNoFnIncludingTypedWithArgumentsLazy : Parser State (Node TypeAnnotation)
+typeAnnotationNoFnIncludingTypedWithArgumentsLazy =
+    Combine.lazy (\() -> typeAnnotationNoFnIncludingTypedWithArguments)
```
jfmengels, line 61: We can also inline these declarations if we want to:
```elm
Combine.lazy (\() -> typeAnnotationNoFnIncludingTypedWithArguments)
  |> Combine.andThen ...
```
(Just mentioning it because I realized it now, by looking at the Elm docs)

src/Combine.elm
```diff
@@ -378,18 +377,13 @@ sepBy1WithoutReverse sep p =
         |> andThen (\first -> manyWithoutReverse [ first ] (sep |> continueWith p))
 
 
-between : String -> String -> Parser state a -> Parser state a
+between : Core.Parser () -> Core.Parser () -> Parser state a -> Parser state a
 between lp rp (Parser p) =
     Parser <|
         \state ->
-            Core.map (\() -> identity) (Core.symbol lp)
+            Core.map (\() -> identity) lp
```
jfmengels, line 384: We could extract this line into a let declaration outside of the lambda

src/Elm/Parser/TypeAnnotation.elm
```diff
@@ -13,26 +13,24 @@ import Parser as Core exposing ((|=))
...
-                Layout.optimisticLayoutWith
-                    (\() -> typeRef)
-                    (\() ->
-                        Combine.oneOf
-                            [ arrowRightToTypeAnnotation
-                                |> Combine.map (\ta -> Node.combine TypeAnnotation.FunctionTypeAnnotation typeRef ta)
-                            , Combine.succeed typeRef
-                            ]
-                    )
+    Combine.map (\ta -> \fromTa -> fromTa ta)
```
jfmengels, line 16: Pretty cool trick!

lue-bird: Sadly so far it doesn't seem to show any improvements for me. I'm in the process of converting most andThens to this style and we'll see if this'll work.

Smaller benchmarks I've made seem to prefer the style where each branch actually parses to a union type of what it knows, then leaves the "caller" to attach their info.
This would lead to a lot of temporary types being created (which to reiterate is faster but might be a bit clutter-y)

lue-bird:
Before I attempt this I'd like to check in with you:
I'd like to try eliminate
```elm
Combine.Parser a =
    ({ indents : List Int
     , comments : List (Node String)
     }
     ->
     Core.Parser
        ( { indents : List Int
          , comments : List (Node String)
          }
          , a
        )
    )
```
in favor of
```elm
ParserWithComments a =
    Core.Parser { comments : Rope (Node String), syntax : a }
```
using indentation handling built into `Core.Parser`

## upsides
- better performance (that's the hope anyway) due to less state and less functions
- no more mixing the two parser kinds, shared helpers between parser representations
- it's pretty easy to follow where information comes from ("no implicit state")
- if we in v8 decide to attach comments in the AST directly, this becomes trivial (more like it would _simplify_ the code)

## downsides
- less control over the indentation state, which means getting the parsed result + indentation in one go isn't possible
- combining comments in the right order is something parser devs have to explicitly get right everytime layout is used

src/Combine.elm
```diff
@@ -239,21 +239,26 @@ maybeIgnore (Parser p) =
...
-    manyWithoutReverse p
-        |> map List.reverse
+many (Parser p) =
+    let
+        manyWithoutReverseStep : ( state, List a ) -> Core.Parser (Core.Step ( state, List a ) ( state, List a ))
+        manyWithoutReverseStep ( oldState, items ) =
+            Core.oneOf
+                [ p oldState
+                    |> Core.map (\( newState, item ) -> Core.Loop ( newState, item :: items ))
+                , Core.succeed (Core.Done ( oldState, List.reverse items ))
```
jfmengels, line 252: I think this should be wrapped in lazy so that we don't compute `List.reverse items` when there are more items left.

src/Elm/Parser/Patterns.elm
```diff
@@ -8,38 +8,57 @@ import Elm.Parser.State exposing (State)
...
+composedWith =
     Combine.maybeIgnore Layout.layout
         |> Combine.continueWith
             (Combine.oneOf
-                [ Core.keyword "as"
+                [ Tokens.asToken
                     |> Combine.fromCoreIgnore Layout.layout
                     |> Combine.continueWithCore
-                        (Node.parserCoreMap (\y -> Node.combine AsPattern x y)
+                        (Node.parserCoreMap PatternComposedWithAs
```
jfmengels, line 24: I wonder whether putting a function (instead of a custom type+pattern matching) would not have been better for performance (similar to the "apply" trick)
```elm
Node.parserCoreMap (\x -> \y -> Node.combine AsPattern x y)
-- Maybe it should be \y -> \x ->
-- I'm not sure yet
```

src/Elm/Parser/Layout.elm
```diff
@@ -2,68 +2,137 @@ module Elm.Parser.Layout exposing
...
+                        ( State.addComments newComments state, () )
+                    )
+        )
+
+
+maybeLayout : Parser State ()
+maybeLayout =
+    coreWhiteSpaceAndComments
+        |> addAsCommentsToState
+        |> verifyLayoutIndent
```
jfmengels: I'm wondering whether we could combine `verifyLayoutIndent` with the previous function to avoid a `Core.map`

(Definitely not for this PR)

src/Elm/Parser/Patterns.elm
jfmengels: I see that later you use the custom type version (https://github.com/stil4m/elm-syntax/pull/233/commits/090a985201d801fdc55d67b5529162cb82ed6033). That ends up being faster I'm guessing?

src/Combine.elm
lue-bird: indeed, fixed. I do wonder why this didn't result in a bigger penalty in performance :thinking:

src/Elm/Parser/Patterns.elm
lue-bird: I did benchmark this. And if the intermediate union type is relatively cheap to construct/reuse it's clearly faster.

But I might be blind for the more heavy intermediates because I do vastly prefer this style over "apply style".
Tho I have also pretty extensively compared these changes using the 100 run benchmark and not found a consistent difference between the styles, so :shrug:

src/Elm/Parser/Layout.elm
```diff
@@ -81,20 +81,18 @@ positivelyIndented : Parser State ()
...
+                        if indent < column then
                             Core.succeed ( state, () )
 
                         else
                             Core.problem "must be positively indented"
-                    )
+                )
+                Core.getCol
+                |= State.currentIndent
+                |> Core.andThen identity
```
jfmengels, line 95: We could even extract the whole core part and use `Combine.fromCore` (if even necessary)

```elm
positivelyIndented : Parser state ()
positivelyIndented =
    Core.map
        (\column ->
            \indent ->
                if indent < column then
                    Core.succeed ()

                else
                    Core.problem "must be positively indented"
        )
        Core.getCol
        |= State.currentIndent
        |> Core.andThen identity
        |> Combine.fromCore
```

jfmengels: Wouldn't my version be faster because we don't redefine the core parser inside a function (the `\state ->` lambda)?

Also, maybe I'm not following which extra map you're thinking of.

src/Elm/Parser/Patterns.elm
jfmengels: Yeah I was thinking that creating a new data structure would be slower than a function, but maybe creating a function (with a lot of closure?) is slower?
But alright, I think the new style is pretty nice :+1:
I loved seeing the journey of the API evolution :smile:

src/Elm/Parser/Layout.elm
lue-bird: in the current we do `Core.succeed ( state, () )`, in the `fromCore` we internally do a `Core.map (\val -> ( state, val ))`.

But might be that constructing this small parser is more expensive than predefining+map but I'm sceptical

jfmengels:
This was amazing @lue-bird, thank you so much for your recent PRs. Performance has been going :rocket: with all these changes



---

> Before I attempt this I'd like to check in with you, I'd like to try eliminate...

Am I understanding correctly that indentation is now solved through the more recent commits? Or would you like to reintroduce `indents` still?

> no more mixing the two parser kinds, less conversion, more shared helpers between parser representations

This would be good, because this optimization has been pretty rough for the project in terms of consistency of code patterns :sweat_smile: 

> So.. does this experiment sound reasonable or is there something to consider which could make this a bad change?

I have little insight into the repercussions of such a change. I'm all for trying it out though. **Go ahead!**

(In the meantime, I'll rebase the v8 branch from #230 on top of `master`)

> accumulating in a Rope adds a bit of overhead and makes e.g. sorted insert (see port documentation comment) harder

For the port documentation comment, that problem will go away with v8, so I don't mind it making the parser slower in the meantime (especially since ports are not that common).

> And should this rather be a separate PR?

Yes. I've merged it anyway, so no choice anymore :smile: 
But yes, explorations that may fail should usually be done in new PRs, so that we don't block work that we know is an improvement from getting merged.

PR #235, opened 2024-07-25
Remove the Operator Expression variant

jfmengels:
Finally! All of the Pratt parser work finally paid off, and we can get rid of this weird node type!


jiegillet (approved):
🎉

miniBill:
HELL YES

PR #236, opened 2024-07-26
Replace combine by core

lue-bird:
as discussed in https://github.com/stil4m/elm-syntax/pull/233#issuecomment-2248836851

- before (taken from "after" of previous PR): 31419.200000000186 ms to do 100 runs => 314.1920000000019 ms/run
- after (purely combine→core): 27615.5 ms to do 100 runs => 276.155 ms/run
- after (+ additional optimizations): 18849.700000001118 ms to do 100 runs => 188.49700000001118 ms/run

So yeah, noticeable but not game-changing.

Code-style-wise, I personally like this new style a lot because of how obvious the information flow is. Though it is undeniably more verbose.


src/Elm/Parser.elm
```diff
@@ -40,14 +39,4 @@ When parsing fails, the result will contain a list of errors indicating what wen
...
-            Ok fileContents
-
-        Err s ->
-            Err s
-
-
-withEnd : Parser State File -> Parser State File
-withEnd p =
-    p |> Combine.ignore end
+    Core.run (file |. Core.end) input
```
jfmengels: Is there a reason that we don't put `Core.end` directly in the definition of `file`? :thinking:

lue-bird: seems reasonable to me.

src/Elm/Syntax/Node.elm
```diff
@@ -44,7 +44,7 @@ empty a =
     Node Range.empty a
 
 
-{-| Combine two nodes, constructing a new node which will have the outer most range of the child nodes
+{-| ParserWithComments two nodes, constructing a new node which will have the outer most range of the child nodes
```
jfmengels, line 47: I think this was unintended :joy:

lue-bird: Thanks for being so thorough with the review! <3 Fixed.

src/ParserWithComments.elm
```diff
@@ -0,0 +1,396 @@
...
+                )
+            )
+
+
+between : Core.Parser () -> Core.Parser () -> ParserWithComments a -> ParserWithComments a
+between lp rp p =
+    (lp
+        |> Parser.Extra.continueWith p
+    )
+        |. rp
```
jfmengels: (Just brainstorming)

These mix of `|.`/`|=` with `|>` require some odd parentheses a bit everywhere. I'm wondering whether we should instead use `|>` pipelines everywhere along with keep/ignore functions.

```elm
-- Maybe with a different name to not
keep = (|=)

foo
   |> keep someCoreParser
```

That way we can keep the constraint that `ignore` should work with `ParserWithComments ()` or `Parser ()`.

lue-bird: `ParserWithComments ()` is not a thing. And we don't want to ignore comments anyway I don't think?
Anyway, could be reasonable. Maybe we just apply continueWith prefix-style to avoid the parens?

jfmengels: Hmm, `Parser.WithComments ()` then? Anyway, what I meant is that prior to this PR, `Combine.ignore` took a `Combine.Parser State ()` as an argument, and that allowed us to make sure that the parser that was being ignored would not compute anything unnecessarily (like `Combine.string` back in 7.3.2). I think that's an attractive property to have if we can keep it.

(Note, I'm still not past the first comment, I'm just brainstorming here :sweat_smile: )

src/Rope.elm
```diff
@@ -0,0 +1,51 @@
...
+
+
+empty : Rope a
+empty =
+    Leaf []
+
+
+one : a -> Rope a
+one onlyElement =
+    Leaf (List.singleton onlyElement)
```
jfmengels, line 19: Unless the compiler inlines `List.singleton`, which I doubt, we should replace this by `Lead [ onlyElement ]`. It's simpler and faster.

src/Rope.elm
```diff
@@ -0,0 +1,51 @@
...
+
+
+fromList : List a -> Rope a
+fromList list =
+    Leaf list
+
+
+empty : Rope a
+empty =
+    Leaf []
```
jfmengels, line 14: Considering how often we will be adding empty leafs, I wonder whether having a separate variant for empty would make sense.
In my mind, we could then potentially pattern match on that and ignore it in `prependTo`/`toListInto`

```elm
type Rope a
    = Leaf (List a)
    | Branch (List (Rope a))
    | Empty () -- So that each variant has the same arity, better for perf.

prependTo : Rope a -> Rope a -> Rope a
prependTo right left =
    case right of
      Empty () -> left
      _ ->
        case left of
            Empty () -> right
            _ -> flatFromList [ left, right ]

toListInto : List a -> Rope a -> List a
toListInto initialAcc rope =
    case rope of
        Empty -> initialAcc
         -- ...
```

I don't know if this will help, but I suspect that we will end up with a rope full with empty ropes (`Branch [ Leaf [], Leaf [], Leaf [], Branch [ Leaf [], Leaf [], Leaf [], Leaf [] ], ... ]`)

@miniBill is this something you have looked into by any chance?

src/Rope.elm
```diff
@@ -0,0 +1,51 @@
...
+
+
+toListInto : List a -> Rope a -> List a
+toListInto initialAcc rope =
+    case rope of
+        Leaf list ->
+            List.foldr (::) initialAcc list
+
+        Branch ropes ->
+            List.foldr (\childRope childAcc -> toListInto childAcc childRope)
```
jfmengels, line 39: This is could potentially create a stack overflow. I wish we had tail-recursion modulo cons here :/

Anyway, I can take a look at it later, but I at the moment don't have a good solution without doing a `List.reverse` at some point.

src/Rope.elm
lue-bird: Funny you would say that. I've even created a port of elm-rope to that representation: https://github.com/lue-bird/elm-rope/tree/explicit-emptiness.
Yeah, I'll test it!

src/Rope.elm
lue-bird: fixed, thanks.

src/Rope.elm
jfmengels: Right, though it could reduce the amount.

Maybe just `Debug.log`ging the rope at the end of parsing `Simplify` could be interesting to see :sweat_smile: 
If it turns out to be mostly filled with comments, then it's already quite good. But my suspicion is that it's very filled with air.

jfmengels: To be explicit, I'm thinking that creating this large but unnecessary data structure has a performance cost. That's a "lot" of memory to allocate and to free up during garbage collection.

lue-bird: Well you were as right as it gets. But let me see if pre-filtering is faster

lue-bird: yep. Is better!

miniBill: I have not in fact looked into that! My guess is that it's not a GC issue because there is a single instance of `empty`, but I could be wrong? It miiiight make sense in `fromList` to `if isEmpty x then empty else Leaf x`?

src/Rope.elm
miniBill: Yeah, unfortunately `foldr` is... annoying. I have a very good `foldl` for `Rope`, but it does require a `reverse` at the end :(

src/Rope.elm
jfmengels: `empty` will indeed always be a single item, with a single allocation. The problem is the `Leaf` and `Batch` that are allocated all the time, and create large lists.

I Debug.logged the resulting rope of the `Simplify.elm` module, and stored it in this Gist: https://gist.githubusercontent.com/jfmengels/c4234361ef4bf14a13b03c849867d3df/raw/44b4084980827a669fdcf3511d5fd13e6992e62c/Rope%2520emptiness.txt
(Spoiler: it's 1.1MB of text, so pretty big)

src/Rope.elm
jfmengels: Maybe it's faster? :sweat_smile:
Honestly, if the rope is large but the resulting list is small, then I think foldl+reverse will be faster.

src/Rope.elm
lue-bird: yall are the best. I would have never believed that we can get parsing down to consistently less than 210 ms/run (with likely potential for even more speed)

src/Rope.elm
```diff
@@ -1,41 +1,80 @@
...
 
 
+prependToLikelyFilled : Rope a -> RopeLikelyFilled a -> Rope a
+prependToLikelyFilled right leftLikelyFilled =
+    case right of
+        Nothing ->
+            Just leftLikelyFilled
+
+        Just rightLikelyFilled ->
+            Just (Branch [ leftLikelyFilled, rightLikelyFilled ])
```
jfmengels, line 33: Maybe we could skip the `Maybe` and pattern match on the contents of `Leaf` or `Branch`?
```elm
prependToLikelyFilled : Rope a -> RopeLikelyFilled a -> Rope a
prependToLikelyFilled right leftLikelyFilled =
    case right of
        Leaf [] ->
            Just leftLikelyFilled

        Branch [] ->
            Just leftLikelyFilled

        _ ->
            Just (Branch [ leftLikelyFilled, right ])
```
I think that could be faster, but I'm unsure.

lue-bird: Hmm. Runs slower.

I've also tried:
- Introduce a `One a` variant → slower
- Introduce a `Branch2 filled filled` → not decisive

now testing the `Empty` variant thing

src/Elm/Parser/Patterns.elm
```diff
@@ -273,10 +299,17 @@ recordPattern =
...
                 (\commentsBeforeElements ->
-                    \elements ->
-                        { comments = Rope.flatFromList [ commentsBeforeElements, elements.comments ]
-                        , syntax = RecordPattern elements.syntax
-                        }
+                    \maybeElements ->
+                        case maybeElements of
+                            Nothing ->
+                                { comments = commentsBeforeElements
+                                , syntax = RecordPattern []
```
jfmengels, line 314: Can be extracted to a top-level variable to avoid reallocating memory.

lue-bird: the `RecordPattern []`? The function?

jfmengels: I meant `emptyRecordPattern = RecordPattern []`

lue-bird: done.

src/Elm/Parser/Expression.elm
```diff
@@ -753,11 +753,11 @@ negationOperation =
                         (Node { start, end } _) =
                             subExpressionResult.syntax
                     in
-                    { comments = subExpressionResult.comments
-                    , syntax =
-                        Node
-                            { start = { row = start.row, column = start.column - 1 }, end = end }
-                            (Negation subExpressionResult.syntax)
+                    { subExpressionResult
```
jfmengels, line 842: A record update expression is slower than creating a new one. A performance tip is usually to try to have the former code rather than the record update.

src/Elm/Parser/Layout.elm
```diff
@@ -37,6 +37,19 @@ nonEmptyWhiteSpaceOrComment =
...
 whiteSpaceAndComments : Parser Comments
 whiteSpaceAndComments =
+    Parser.oneOf
+        [ Parser.keyword " "
+            |> Parser.andThen (\() -> whiteSpaceAndCommentsLoop)
+        , Parser.token " "
+            |> Parser.map (\() -> Rope.empty)
+
+        -- fallback if the hacky shortcuts don't commit
+        , whiteSpaceAndCommentsLoop
```
jfmengels: Could you explain the trick here? I'm not sure I get it.

src/Elm/Parser/Layout.elm
```diff
@@ -18,37 +18,58 @@ import Rope exposing (Rope)
 import Set
 
 
-nonEmptyWhiteSpaceOrComment : Parser.Parser (Maybe (Node String))
-nonEmptyWhiteSpaceOrComment =
+{-| Nothing if there can't be more whitespace/comments ahead,
+otherwise Just with the next comment node
```
jfmengels, line 20: This documentation comment is out of date in the last state of the PR

src/Elm/Parser/Expression.elm
lue-bird: Thought there was an optimization for that in eol2. Fixed.

src/Elm/Parser/Layout.elm
lue-bird: The trick was: Since a single space is by far the most common maybeLayout, we
- first check whether there is not text/numbers/... after the space using `keyword`. If there isn't, we can assume (if elm-format is used) that there will be more whitespace following that (that's obviously not a general rule but a common rule because of how common references/variables are)
- If there is, we (if elm-format is used) very likely have a single space which we can now reach and parse very fast

These are very fragile/stupid heuristics that did make a small dent. I'm happy the new thing is faster without.

src/Elm/Parser/Layout.elm
lue-bird: thanks, fixed.

src/Rope.elm
lue-bird: completely switching out `Branch (List filled)` for `Branch2 filled filled` and `Leaf (List a)` for `Leaf a` seemed to have done the trick, since a branch in the previous version pretty much never had more than 2 filleds per branch.

A nice extra is that it's now not a "likely filled" but a typesafe "filled".

src/Elm/Parser/Expression.elm
jfmengels: I think you're thinking of https://blogg.bekk.no/improving-the-performance-of-record-updates-cb34cb6d4451

I think it's better in eol2 compared to without, but I would imagine it's still slower than a plain record definition (if the Elm compiler used opportunistic mutation, maybe then...)

jfmengels:
Well, this has been amazing once more. Another 50%+ performance increase :sweat_smile:
(though I wonder how it performs without eol2)

Great chances @lue-bird, thank you so much! I'm glad you started this experiment, and I hope you had fun! :wink:

PR #239, opened 2024-07-31
Use ParserWithComments.until more

jfmengels:
Am I doing this right?

### DIFF
--- a/src/Elm/Parser/Expose.elm
+++ b/src/Elm/Parser/Expose.elm
@@ -46,7 +46,6 @@ exposeListWith =
             )
     )
         |= exposingListInner
-        |. Tokens.parensEnd
 
 
 exposingListInner : Parser (WithComments Exposing)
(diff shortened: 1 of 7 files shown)


lue-bird:
I'm sad you did all this work. I expect this change to make things slower for a simple reason:
The only reason `until` can be faster than `many` sometimes is that it has the ability to very cheaply shortcut to the end instead of doing a costly extra check at the end.
This improvement is totally useless when every element to check starts with a simple token and therefore is very cheap itself. And since enumerations contain much more than 1 element, we just introduced a whole bunch of extra checks for the end token.

The only thing that could improve things is your new declarations parser (but the effect is likely minimal/negative for longer files)

jfmengels:
> I'm sad you did all this work.

No worries, it only took me a few minutes. If I'm learning something it's okay :+1: 

> The only reason until can be faster than many sometimes is that it has the ability to very cheaply shortcut to the end instead of doing a costly extra check at the end.

I thought the benefit of `until` was that in practice it did `Parser.map Done someParser` instead of `Parser.succeed (\() -> Done () |. someParser` (or something very similar), skipping on a few operations.

lue-bird:
Well the `map` is usually _after_ the `Token.comma |> Parser.Extra.continueWith ..here..` so it shouldn't impact performance

lue-bird:
That's actually a good observation! We can definitely introduce a helper that's like `until` except the `Loop` and `Done` branch swapped :+1: 
So feel free to reopen!

jfmengels:
I'm not sure what I'm doing, so I'll keep it closed. Feel free to re-use these commits or ideas if you feel like they're worth exploring.

PR #243, opened 2024-08-02
Add failing test when file contains a large amount of comments

jfmengels:
This creates a stack overflow. I am not sure how to fix it, but I have created a regression test, should anyone want to look into this :pray:


lue-bird:
for reference, I tried replacing our maybeLayout with a `loop` and is got ~10ms slower.
What I think we could do though is _only_ looping once we encounter a single-line comment (might be a bit rough though)

jfmengels:
I think that would make sense, although I'm not sure what you mean by rough.

PR #244, opened 2024-08-02
Make arguments for FunctionCall nonempty

jfmengels:


jfmengels:
(unless someone wants to review changes like these) I'll be merging these changes rather fast if they go in the right direction for what was decided for `v8`. Please do comment on these PRs even after merging (or open new issues) if you think something is off. We'll probably have to rebase the `breaking-changes-v8` branch a lot anyway so I'm happy with removing and re-doing commits anyway.

PR #247, opened 2024-08-03
Move the operator argument for Operation and wrap it in Node

jfmengels:
Wrapping it in a Node makes the position available, which I know we've used workarounds for in some `elm-review` rules.

Moving the operator is meant to make the data resemble the AST more (because it's `a + b`, not `+ a b`)

I'm not sure if we can improve the parser part more.

### DIFF
--- a/src/Elm/Parser/Expression.elm
+++ b/src/Elm/Parser/Expression.elm
@@ -1194,7 +1194,7 @@ applyExtensionRight extensionRight ((Node { start } left) as leftNode) =
                     extendRightOperation.expression
             in
             Node { start = start, end = end }
-                (Operation extendRightOperation.symbol leftNode right)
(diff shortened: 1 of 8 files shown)


src/Elm/Parser/Expression.elm
```diff
@@ -1268,35 +1268,46 @@ computeAbovePrecedence currentPrecedence =
...
+        (\() ->
+            \( endRow, endColumn ) ->
+                Node
+                    { start = { row = endRow, column = endColumn - length }
+                    , end = { row = endRow, column = endColumn }
+                    }
+                    symbol
         )
+        (Parser.symbol symbol)
+        |= Parser.getPosition
```
jfmengels, line 1293: (I'm slightly happy with this trick that avoids an eager `getPosition`)

PR #252, opened 2024-08-04
Speed up Node parsers

jfmengels:
This only splits up the arguments into several lambdas.


jfmengels:
Irrelevant now with the new custom parser.

PR #253, opened 2024-08-05
custom parser

lue-bird:
Contains tiny improvements to the old parser
followed by the more important conversion to a custom parser.
- before (taken from last PR): 16077.300000000745 ms to do 100 runs => 160.77300000000744 ms/run
- after initial optimizations: 15633.79999999702 ms to do 100 runs => 156.3379999999702 ms/run
- after conversion to custom parser: 12211.5 ms to do 100 runs => 122.115 ms/run

The new parser
- saves a lot of mapping after e.g. `symbol`
- saves many combinators/andThens, e.g. for ranges/locations, layout indentation checks and lookahead/lookbehind
- replaces keep/ignore pipelines to `mapN`

This should still preserve type compatibility of the package interface since we re-use `Parser.Problem`.
In terms of performance, this change opens the door for a lot of optimizations that were locked off before (e.g. removing the `context` field from the state or returning `{ row, column }` instead of a tuple)


src/Elm/Parser/Comments.elm
```diff
@@ -30,12 +31,11 @@ multilineCommentString : Parser.Parser String
...
         [ Parser.symbol "{-|"
-            |> Parser.backtrackable
-            |> Parser.map (\() -> problemUnexpected)
+            |> Parser.Extra.continueWith problemUnexpected
         , Parser.multiComment "{-" "-}" Nestable
-            |> Parser.mapChompedString (\comment () -> Parser.succeed comment)
+            |> Parser.getChompedString
         ]
-        |> Parser.andThen identity
+        |> Parser.backtrackable
```
jfmengels: Does this make the whole `{- -}` backtrackable? I'm guessing that's only a performance problem if there is no closing `-}`?

src/Elm/Parser/Expression.elm
```diff
@@ -916,21 +916,15 @@ negationOperation =
...
+                , Parser.succeed ()
                 ]
             )
-        |> Parser.andThen identity
-
-
-problemNegationThenSpace : Parser.Parser a
-problemNegationThenSpace =
-    Parser.problem "negation sign cannot be followed by a space"
+        |> Parser.backtrackable
```
jfmengels: Does this not make backtrackable operate on a "large" section of the source?

src/Elm/Parser/Comments.elm
lue-bird: removed in favor of lookahead now

src/Elm/Parser/Expression.elm
lue-bird: replaced by smarter sub expression parser now (one without initial layout)

lue-bird:
Wow, I'm surprised this bug hasn't happened for folks.
If we have e.g.
```elm
function WithoutArgs anotherArgument =
    0
```
(same with let function arguments)
It previously parsed as `NamedPattern "WithoutArgs" [ VarPattern "anotherArgument" ]`.
The fix now also makes `as` pattern usage more robust while speeding it up because there's no need to check for composing.

src/Elm/Parser/Tokens.elm
```diff
@@ -163,7 +164,7 @@ functionName =
         { inner =
             \c ->
                 -- checking for these common ranges early is much faster
-                charIsAlphaNumFast c || c == '_' || Unicode.isAlphaNum c
+                Char.Extra.isAlphaNumFast c || c == '_' || Unicode.isAlphaNum c
```
jfmengels, line 166: Should we try to use a version of `Unicode.isAlphaNum` that doesn't check the same ranges as `Char.Extra.isAlphaNumFast`?

src/CustomParser/Advanced.elm
```diff
@@ -1321,31 +1321,34 @@ isSubChar predicate offset string =
...
             String.slice offset (offset + 1) string
     in
-    case actualChar of
-        "" ->
-            -1
+    if charStringIsUtf16HighSurrogate actualChar then
+        -- String.all iterates over code points (so here just one Char)
+        if String.all predicate (String.slice offset (offset + 2) string) then
+            case actualChar of
+                "" ->
```
jfmengels, line 1406: I don't thin `actualChar` can be empty, can it?

src/CustomParser/Advanced.elm
```diff
@@ -1228,6 +1228,54 @@ changeIndent newIndent s =
...
+mapWithStartPosition :
+    ({ row : Int, column : Int } -> a -> b)
+    -> Parser x a
+    -> Parser x b
+mapWithStartPosition combineStartAndResult (Parser parse) =
+    Parser
+        (\s0 ->
+            case parse s0 of
+                Good p a s1 ->
+                    Good p (combineStartAndResult { row = s0.row, column = s0.col } a) s1
```
jfmengels, line 1295: :heart:

src/CustomParser/Advanced.elm
lue-bird: charStringIsUtf16HighSurrogate does not check for emptiness, it uses `String.all` to be fast

jfmengels: I was more thinking in the sense that `String.slice offset (offset + 1) would never be empty, but I guess this could be the end of the file. Alright.

src/CustomParser/Advanced.elm
```diff
@@ -1058,6 +1058,43 @@ mapChompedString func (Parser parse) =
         )
 
 
+chompIfFollowedBy : (Char -> Bool) -> x -> Parser x a -> Parser x a
+chompIfFollowedBy isGood expecting (Parser parseNext) =
```
jfmengels: Took me a bit of time to understand this one. I read it as "chomp, if followed by" so I thought it was doing look ahead but couldn't understand where :joy: 
(Leave it as is, it's fine, unless you have a better name in mind)

jfmengels:
This has been another great series of changes. Great work @lue-bird :heart:

PR #256, opened 2024-08-17
WIP: Smaller optimizations for custom parser

lue-bird:
- before: 7583.10000000149 ms to do 100 runs => 75.8310000000149 ms/run
- after: 2784.2999999970198 ms to do 100 runs => 27.842999999970196 ms/run

(my old laptop broke down, so both of these are from my new one)

One change I really like is that now e.g. `(\a -> { x = 3 }).x` doesn't parse.

## considering
eliminate all mapWithRange

? try eliminate the last few oneOfs, also consider for infixes

? disallow variant/record type alias constructor with record access

(somewhat high effort, might not gain much) rewrite some chompWhiles and variables to avoid utf16 surrogate check
for functionName and typeName


src/Elm/Parser/Imports.elm
```diff
@@ -3,113 +3,92 @@ module Elm.Parser.Imports exposing (importDefinition)
...
-                                                    Nothing ->
-                                                        Rope.empty
+                                Nothing ->
+                                    modRange
+            in
+            { comments =
+                commentsAfterImport
+                    |> Rope.prependTo commentsAfterModuleName
+                    |> Rope.prependTo
+                        (case maybeModuleAlias of
```
jfmengels, line 46: Slightly faster would likely be to do
```elm
|> (case maybeModuleAlias of
    Nothing -> identity
    Just moduleAliasValue -> Rope.prependTo moduleAliasValue.comments
   )
```
but I think it would be even faster to introduce a variable:
```elm
let
  commentsUntilAfterModuleName =
      commentsAfterImport
                    |> Rope.prependTo commentsAfterModuleName
in
(case maybeModuleAlias of
    Nothing -> commentsUntilAfterModuleName
    Just moduleAliasValue -> Rope.prependTo moduleAliasValue.comments commentsUntilAfterModuleName
)
  |> Rope.prependTo ...
```

(This pattern is used in a number of places and files, so if one of these is chosen, this can be applied elsewhere as well)

jfmengels:
I see a lot of `String.slice` to get a single character. That is apparently quite slow (https://jsbench.me/oum0f8qyn3/1) compared to a simple `string[index]`. I think that if we make a dedicated function for that (getting a single argument), then in `elm-review` I can replace it by a function that uses `string[index]`. Hopefully that will make things a lot faster. To be benchmarked as well obviously.

src/Elm/Parser/Imports.elm
lue-bird: went with B as that should be a tiny bit faster indeed (without seeing any perf improvements, but that's expected)

lue-bird:
I tried to run the regression script on the broken file with the many comments and it succeeded. Since I'm not fully sure I used it correctly, are you seeing the same?

And would there be other blockers you'd like to get fixed?

jfmengels:
Seems like the branch is broken since the last push, there's non-compiling code like this:
```elm
mapWithStartAndEndPosition :
    (Location -> a -> Location -> b)
    -> Parser a
    -> Parser b
mapWithRange combineStartAndResult (Parser parse) =
  -- ...more code
```

> And would there be other blockers you'd like to get fixed?

If no regressions are found (I'll try to run the regression script when I can), then there's nothing except except this regression: https://github.com/stil4m/elm-syntax/pull/243/commits/2af7c66c37f95a8781a03c3df2216785d380a359

PR #257, opened 2024-08-30
Smaller optimizations for custom parser since before incorrect force push

lue-bird:
starts at #256 without the incorrect rebase into force-push.

This branch rebased, accepting all commits until `git merge-base smaller-optimizations-for-custom-parser-since-before-incorrect-force-push upstream/master` (which is `a06ca8a68744204edd00d38de4c29970dfa85b44`),
followed by a `git rebase --onto upstream/master <the merge base>` and finally a force-push.

In case this overwrites some changes you've done since or is otherwise problematic, please comment.

Before, I tried `rebase -Xtheirs` and `rebase -X ours` but these seem to have included commits from upstream/master that messed up future history from my branch.

----------
> actual PR content

- before: 7583.10000000149 ms to do 100 runs => 75.8310000000149 ms/run
- after: 2761.39999999851 ms to do 100 runs => 27.613999999985097 ms/run

(my old laptop broke down, so both of these are from my new one)

One change I really like is that now e.g. `(\a -> { x = 3 }).x` doesn't parse.


src/Elm/Parser/Declarations.elm
```diff
@@ -592,71 +611,66 @@ customTypeDefinitionAfterDocumentationAfterTypePrefix =
...
-                                        let
-                                            (Node headVariantRange _) =
-                                                typeDeclarationAfterDocumentation.headVariant
-                                        in
-                                        headVariantRange.end
-                        in
-                        { comments = allComments
-                        , declaration =
-                            Declaration.CustomTypeDeclaration
+    ParserFast.map2WithStartLocation
```
jfmengels, line 614: We do some computations here to get the `end` position. Would it be faster to use `map2WithRange`?

lue-bird: That would mean unnecessary computation and application for all other declarations. A case-of is pretty cheap anyway and has the added benefit the we could easily get rid of the backtracking for the last variant.

Issue #258, opened 2024-09-10
Binary operator parsing precedence bug


wolfadex:
**Describe the bug**
`a + b /= 0` is marked as invalid but the Elm compiler and elm-test don't have an issue with it. By wrapping the math in parens, like `(a + b) /= 0`, then it parses fine.  This also applies to `a + b + c /= 0` where wrapping like `(a + b  +c ) /= 0` fixes it.

**Additional context**
This was originally filed in elm-review https://github.com/jfmengels/elm-review/issues/177 as that's where the bug was found. Elm Slack thread for discussion reference https://elmlang.slack.com/archives/C010RT4D1PT/p1725977809992319.

jfmengels:
Fixed in `7.3.6`

PR #259, opened 2024-09-10
Successfully parse 'a + b /= c'

jfmengels:
Fixes https://github.com/stil4m/elm-syntax/issues/258

We usually attempt to parse '/=' before we parse '/', which parses successfully.
But because in this expression we previously encountered a '+', we're at a precedence leve where we are only evaluating operators with a higher precedence than '/='. So we don't check for '/=', and then we see a '/' which commits, and then we get stuck because we don't expect a '='.

This fixes the issue by checking, for the '/' operator, that it is not followed by a '='. Ideally we should only do this if the operator
precedence is higher than the precendence for '/=', but that data is not available for the parser.

Similarly, this fixes the parsing of 'a && b <| c' which has the same problem.

cc @lue-bird, let me know if I missed something or if there is a better or more performant way to do this.

---

On top of this, I also made a small improvement to the printing of `DeadEnd`s in the tests.


jfmengels:
Superseeded in #260.

PR #261, opened 2024-09-11
correct float handling of the custom parser

lue-bird:
I'm not getting regression notices anymore. Issue was bad float handling in the new custom parser replacement


jfmengels:
I don't see any regressions anymore either. I'll release this now!
(and the regressions script tell me this is a 50% performance improvement :rocket: )

PR #264, opened 2024-10-08
Fix incorrect range for imports with alias and exposing

jfmengels:


lue-bird (approved):
Yep that works. (when I experimented in a branch, [only casing on alias and exposing once](https://github.com/stil4m/elm-syntax/commit/a100137a82634dbd490116464a40b49d81c39309) was a bit faster but doesn't really matter)

jfmengels:
Let's take your version if it's faster then. Do you mind making a PR?

lishaduck:
Superseded by #264.

PR #266, opened 2024-10-31
Fix pattern :: comment order

lue-bird:
Noticed that incorrect sorting of cons pattern comments broke my tests.


jfmengels:
Nice! I've added a test in d265c2e8107615fdbbdd8027e0bbcae85de8612b to catch future regressions.

jfmengels:
Released in 7.3.7.

PR #267, opened 2024-11-22
Remove superfluous exposure of imported `Problem` type to top-level

Viir:
Looks like that `exposing (Problem(..))` was introduced with commit <https://github.com/stil4m/elm-syntax/commit/61c89e7cc0d7c84da0108edd049323fe3b72707d> but that seems more like an accidental automated transformation than intended.

### DIFF
--- a/src/ParserFast.elm
+++ b/src/ParserFast.elm
@@ -125,7 +125,7 @@ Once a path is chosen, it does not come back and try the others.
 
 import Char.Extra
 import Elm.Syntax.Range exposing (Location, Range)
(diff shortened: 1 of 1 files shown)


lue-bird (approved):
Thanks. I guess this means there's a false negative in elm-review-unused

jfmengels:
Yes, I'm looking into it. It's a surprisingly tricky one, but I'm almost there (I have a fix, but can't reproduce it in a unit test. Turns out, there's also a false positive around this.
Thank you @Viir :blush:

PR #268, opened 2024-12-07
Optimize comparisons to chars

miniBill:
image

### DIFF
--- a/src/Elm/Parser/Expose.elm
+++ b/src/Elm/Parser/Expose.elm
@@ -79,8 +79,34 @@ infixExpose =
         )
         (ParserFast.symbolFollowedBy "("
             (ParserFast.ifFollowedByWhileWithoutLinebreak
-                (\c -> c /= ')' && c /= '\n' && c /= ' ')
-                (\c -> c /= ')' && c /= '\n' && c /= ' ')
+                (\c ->
+                    case c of
+                        ')' ->
+                            False
+
+                        '\n' ->
+                            False
+
+                        ' ' ->
+                            False
+
+                        _ ->
+                            True
+                )
+                (\c ->
+                    case c of
+                        ')' ->
+                            False
+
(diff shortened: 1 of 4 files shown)


jfmengels:
This is super weird, because I can reproduce tests being many time faster, but when I run the performance script

```bash
./find-regressions.sh --no-check ~/.elm/0.19.1/packages/
```

then performance is about the same, maybe even worse
```
Published: avg 1.5168321927699087
Current:   avg 1.5336765493284654
Diff: -1.0982991534904896% faster
```

src/Elm/Parser/Expose.elm
```diff
@@ -79,8 +79,34 @@ infixExpose =
...
+                            False
+
+                        '\n' ->
+                            False
+
+                        ' ' ->
+                            False
+
+                        _ ->
+                            True
```
jfmengels, line 108: Maybe it's time we extract this to a named function :sweat_smile:

lue-bird:
Using the code in benchmark/ I also don't notice any improvement

jfmengels:
Right, I agree I think it's a difference between having `--optimize` and not having it. And this change manually gets us closer to that. It's impressive that these tiny changes have so much impact on performance in non-optimized though.

Even if the final performance does not get improved (which it might, just a bit hard to measure), it makes tests run much faster, and that's a nice win for the people working on this project!

Thank you @miniBill :heart:

jfmengels:
@lue-bird do you think this is worth releasing? :thinking:

jfmengels:
This is also going to make `elm-review` faster when run with `--debug`, as well as `elm-review` rule tests :heart:

lue-bird:
I applied the same changes for elm-syntax-format's lenient parser and got a 5-10% improvement (_with_ eol2 applied), that's amazing.

PR #271, opened 2025-02-12
fix negation not being parsed after a comma

lue-bird:
Fixes #270 and adds a test for it.


jfmengels:
Looks great!
I think there are actually a number of other characters this could appear after that we don't support. I'm making up a PR right now.

PR #272, opened 2025-02-12
Support more characters directly before negation

jfmengels:
Follow up to #271 

I went through https://lucamug.github.io/elm-cheat-sheet/elm-cheat-sheet.color.pdf and found a bunch more characters that should be allowed before `-`, which I've tested in the REPL. I've written tests for those. I've also added a few regressions tests for patterns that should not be supported (mostly things like `<|-` which the compiler understands as an unknown operator).

I have not succeeded in supporting all of them though. I'd love help with this.

### DIFF
--- a/tests/Elm/Parser/ExpressionTests.elm
+++ b/tests/Elm/Parser/ExpressionTests.elm
@@ -3,7 +3,7 @@ module Elm.Parser.ExpressionTests exposing (all)
 import Elm.Parser.Declarations
 import Elm.Parser.Expression exposing (expression)
 import Elm.Parser.ParserWithCommentsTestUtil as ParserWithCommentsUtil
-import Elm.Syntax.Expression exposing (Expression(..))
+import Elm.Syntax.Expression exposing (Expression(..), LetDeclaration(..))
 import Elm.Syntax.Infix as Infix exposing (InfixDirection(..))
 import Elm.Syntax.Node exposing (Node(..))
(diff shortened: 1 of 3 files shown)


lue-bird (approved):

tests/Elm/Parser/ExpressionTests.elm
```diff
@@ -590,6 +601,116 @@ all =
...
+                                        (Negation
+                                            (Node { start = { row = 1, column = 5 }, end = { row = 1, column = 6 } } (FunctionOrValue [] "x"))
+                                        )
+                                    )
+                                ]
+                            )
+                        )
+        , test "negated expression after ==" <|
+            \() ->
+                "a==-x"
```
lue-bird, line 622: in theory this one should not parse if we want to be that strict
```
-- UNKNOWN OPERATOR ------------------------------------------------------- REPL

I do not recognize the (==-) operator.

1|   a==-x
      ^^^
Is there an `import` and `exposing` entry for it? Maybe you want (==) or (-)
instead?
```

jfmengels: Oops, right.

And I agree that the current system is not enough for `in`.

jfmengels:
@lue-bird I don't think I'll merge this unless tests pass. Feel free to push to this branch (or create a new branch/PR on top of this one) if you'd like to continue the work.

PR #275, opened 2025-06-10
Fix infinite recursion on end of file after quotes or [glsl|

### LINKED ISSUE #273: Infinite loop in the parser
The following (invalid!) Elm code produces an infinite loop in the parser

```elm
module A exposing (..)

a = "
```

lue-bird:
fixes #273 for single quote, single double quote, triple double quote and `[glsl|`.

I have not verified performance but it should be pretty much equal.


jfmengels:
@miniBill: "elm-syntax has this one bug"
@lue-bird *fixes 4 bugs*

Great work as always :smile:

